PR #56, opened 2024-02-03
[feat] refactor workflow to publish package

AndyBoWu:


.github/workflows/publish-package.yaml
```diff
@@ -0,0 +1,104 @@
+name: Publish to npm, Tag and GH Release
+
+on:
+  push:
+    branches:
+      - master
```
edisonz0718, line 6: I think we don't need to add 'master' here. Only main trigger the push

edisonz0718 (approved):

PR #60, opened 2024-02-08
Improve Dispute module functions

DonFungible:


packages/core-sdk/src/resources/dispute.ts
```diff
@@ -19,102 +22,240 @@ export class DisputeClient {
...
+          request.linkToDisputeEvidence,
+          stringToHex(request.targetTag, { size: 32 }),
+          request.calldata ? request.calldata : "0x",
+        ],
       });
 
       const txHash = await this.wallet.writeContract(call);
 
+      if (request.txOptions?.waitForTransaction) {
+        await waitTx(this.rpcClient, txHash);
```
edisonz0718, line 58: I think we should return dispute id in the response, so that users can use dispute id to query API to get dispute information

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -19,102 +22,240 @@ export class DisputeClient {
...
+   * @emits DisputeCancelled (_disputeId, _data);
+   */
+  public async cancelDispute(request: CancelDisputeRequest): Promise<CancelDisputeResponse> {
     try {
       const { request: call } = await this.rpcClient.simulateContract({
         ...DisputeModuleConfig,
-        functionName: "whitelistArbitrationRelayer",
-        args: [request.arbitrationPolicy, request.arbitrationRelayer, request.allowed],
+        functionName: "cancelDispute",
+        args: [BigInt(request.disputeId), request.calldata ? request.calldata : "0x"],
```
edisonz0718, line 124: I usually just do: 
 request.calldata || "0x"

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -19,102 +22,240 @@ export class DisputeClient {
...
-    } catch (error) {
-      handleError(error, "Failed to raise dispute");
-    }
+  /**
+   * Reads from the contract's disputes mapping storage variable.
+   * @param disputeId string | number | bigint, The ID of the dispute to be read.
+   * @returns The address of the Dispute module that disputeId maps to
+   * @calls mapping(uint256 disputeId => Dispute dispute) public disputes;
+   */
+  public async readDisputes({ disputeId }: { disputeId: number | string | bigint }) {
```
edisonz0718, line 182: I think if we want to support read functions, we need to define and export the return types. Otherwise users have to figure out the types and define the types in their application

edisonz0718: I feel that based on the user story we created, these read functions are not important, except readDisputes, which is supported by API. 
It's better for us to keep the SDK lean and make it easy for users by just saying SDK is for writing and API is for reading.

packages/core-sdk/src/abi/sdkEntities.json
```diff
@@ -1,34 +1,39 @@
 [
   "ipAsset",
   "execute",
-
```
edisonz0718, line 4: I organized this file by different resource types and use one line to separate them. Also delete a few unused ones. All events are put in the last.  
Can you just add the functions you need for the dispute? Otherwise I have to redo and make it work again

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -16,59 +23,101 @@ describe.skip("Dispute Functions", () => {
     senderAddress = config.account.address;
     client = StoryClient.newClient(config);
   });
-  /*
-  it("should be able to whitelist dispute tags and wait for transaction", async () => {
-    const response = await expect(
-      client.dispute.whitelistDisputeTags({
-        tag: "testTag",
-        allowed: true,
+  describe("[Write Functions] Should be able to", async function () {
```
edisonz0718, line 26: All write functions are skipped right now. Are they working?

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -1,56 +1,56 @@
...
+  data?: `0x${string}`;
 };
 
-export type RaiseDisputeRequest = {
-  targetIpId: `0x${string}`;
-  arbitrationPolicy: `0x${string}`;
-  linkToDisputeEvidence: string;
-  targetTag: string;
+export type CancelDisputeRequest = {
+  disputeId: number | string | bigint;
```
edisonz0718, line 40: I think we just use either number or string here. Otherwise users have to figure out which types to use.

packages/core-sdk/test/integration/dispute.test.ts
DonFungible: Doesn't work unless we have a working ArbitrationPolicy fixture

DonFungible: Also, setDisputeJudgement won't work unless we set up the whitelisted Arbitration relayer contract (or whatever it's called)

packages/core-sdk/src/types/resources/dispute.ts
DonFungible: I think bigint is ok here. If a user uses ethers/wagmi to read directly from the contract, that output is bigint type.

edisonz0718 (approved):

PR #69, opened 2024-02-18
Initialize react hooks

DonFungible:


DonFungible:
Deprecating this for the time being

PR #101, opened 2024-03-11
Upgrade the viem version to ^2.7.0

bonnie57:
Upgrade the `viem` version to ^2.7.0 and fix the type error.

Test results are available in the following snapshot.

- Unit tests
image

- integration Tests
image


DonFungible (approved):
lgtm

PR #109, opened 2024-03-22
fix: story config account type

jacob-tucker:


DonFungible:
I don't see a package version increase

PR #114, opened 2024-03-27
Implement multiple chains

bonnie57:
## Description
This PR implements multiple chains because of supporting `renaissance` chain, includes：
- 1. Implement multiple chains including `renaissance` and `sepolia`
- 2. Add integration test about mintNFT

## Test Result
I have attached snapshots of the test results.
image
image


packages/core-sdk/src/types/config.ts
```diff
@@ -12,7 +12,9 @@ export type SupportedChainIds =
   | "mainnet"
   | "80001"
   | "mumbai"
-  | "polygonMumbai";
+  | "polygonMumbai"
+  | "1513"
+  | "storyNetwork";
```
edisonz0718, line 17: Not sure about the name here: I think we can call it storyTestnet first

packages/core-sdk/src/utils/env.ts
```diff
@@ -22,3 +24,48 @@ export const sepolia = {
   RoyaltyPolicyLAP: "0x16eF58e959522727588921A92e9084d36E5d3855",
   TokenWithdrawalModule: "0x66738a5722e9Cd91B694bA39146CFEcE95457F04",
 };
+
+const storyNetwork = {
```
edisonz0718, line 28: rename it to storyTestnet

packages/core-sdk/src/types/config.ts
```diff
@@ -12,7 +12,9 @@ export type SupportedChainIds =
   | "mainnet"
   | "80001"
   | "mumbai"
-  | "polygonMumbai";
+  | "polygonMumbai"
```
edisonz0718, line 15: We can delete mumbai and polygonMumbai  and mainnet related config. We are not supporting them in our next sdk release

packages/core-sdk/src/utils/utils.ts
```diff
@@ -155,7 +156,45 @@ export function chainStringToViemChain(chainId: SupportedChainIds): Chain {
     case "mumbai":
     case "polygonMumbai":
       return polygonMumbai;
+    case "1513":
+    case "storyNetwork":
+      return storyNetwork;
```
edisonz0718, line 161: same. rename to storyTestnet

PR #162, opened 2024-05-04
SPG Resource & Test Setup

jdubpark:
## Description

1. Add periphery repo ABI & wagmi generation
2. Add the new resource, SPG Client, for SPG-specific functions (currently only `createCollection`)
3. Modify existing SPG test to create a collection before SPG-based IPA tests -> make the initial test case work.

## Test Plan 

1. In `ipAsset.test.ts`, create an SPG NFT collection using the SPG Client before executing SPG-related IPA tests.


packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -209,7 +209,7 @@ export class IPAssetClient {
    *   @param request.txOptions [Optional] The transaction options.
    * @returns A Promise that resolves to an object containing the transaction hash.
    */
-  private async mintAndRegisterIpAndAttachPilTerms(
+  public async mintAndRegisterIpAndAttachPilTerms(
```
DonFungible, line 212: Throwing out some other suggestions:

- `createIpAssetWithPilTerms`
- `mintIpAssetWithPilTerms`

packages/core-sdk/test/integration/ipAsset.test.ts
```diff
@@ -104,39 +105,36 @@ describe("IP Asset Functions ", () => {
...
-  //       },
-  //     });
-  //     console.log("txHash: ", txHash);
-  //     expect(txHash).to.be.a("string").and.not.empty;
-  //   });
-  // });
+    it("should not throw error when mint and register ip and attach pil terms", async () => {
+      const txHash = await client.ipAsset.mintAndRegisterIpAndAttachPilTerms({
+        nftContract,
+        pilType: PIL_TYPE.NON_COMMERCIAL_REMIX,
```
DonFungible, line 130: should test for all PIL types

.gitmodules
```diff
@@ -1,4 +1,8 @@
-[submodule "packages/core-sdk/protocol-contracts"]
```
edisonz0718, line 1: Since we have the wagmi generator, this part is not really needed I think. But it can be used as a backup when wagmi generator doesn't work. Just FYI

packages/core-sdk/src/client.ts
```diff
@@ -32,6 +33,8 @@ export class StoryClient {
   private _dispute: DisputeClient | null = null;
   private _ipAccount: IPAccountClient | null = null;
   private _royalty: RoyaltyClient | null = null;
+  private _spg: SPGClient | null = null;
```
edisonz0718, line 36: I think we can rename this to NftClient to make it more user friendly. It can host all nft collection related functions that ties to SPG

packages/core-sdk/src/resources/spg.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      if (request.mintCost > 0n && !isAddress(request.mintToken || "")) {
+        throw new Error("Invalid mint token address, mint cost is greater than 0.");
+      }
+
+      const txHash = await this.spgClient.createCollection({
+        name: request.name,
+        symbol: request.symbol,
+        maxSupply: request.maxSupply,
+        mintCost: request.mintCost,
```
edisonz0718, line 47: I think mintCost can be optional too.

DonFungible: Yes, and i think maxSupply should be optional too, in the case of open edition collections

jdubpark: Resolved 7ee307b

packages/core-sdk/src/resources/ipAsset.ts
jdubpark: Resolved e8b0cbb

packages/core-sdk/test/integration/ipAsset.test.ts
jdubpark: Resolved 3bc7b77

packages/core-sdk/src/client.ts
jdubpark: I see your point, but I think we should keep it as SPG since it might also deal with other functionalities in the future. Currently, it's a wrapper for core functions and an NFT collection manager. 

I also think it's better to move all SPG-related functions into SPG since they're an abstraction layer above the core calls that existing resources make (e.g. IPAsset resource calls core functions vs. SPG resource calls SPG that calls core functions).

DonFungible (approved):
lgtm

packages/core-sdk/src/client.ts
jdubpark: Following the discussion, we are switching NftClient to maintain the data type -> resource relationship. Resolved 24b51df

Issue #204, opened 2024-05-30
Error when adding metadata in `.register` function with metamask


jacob-tucker:
**Describe the bug**
If I use Metamask to sign transaction, and I add metadata to the `.register` function, I get the following error: `Error: Failed to register IP: The account does not support signTypedData, Please use a local account.`

If I pass no metadata, it works.
Other functions also work with metamask.
<img width="376" alt="Screenshot 2024-05-29 at 8 41 49 PM" src="https://github.com/storyprotocol/sdk/assets/15198786/215de841-361d-42ce-addd-58ceadfe142e">

bonnie57:
Hi @jacob-tucker ,
Thank you for raising the issue you've encountered. Upon investigation, it seems that the mismatch lies between the account object type obtained from `window.ethereum.request` and the `LocalAccount` object that the SDK expects.

The account details we retrieve:
```
const [account]: [Address] = await window.ethereum.request({
  method: "eth_requestAccounts",
});
```
The account is the wallet address as a string, which differs from the `LocalAccount` object required by the SDK for operations like `signTypedData`. The SDK needs a `LocalAccount` for `signature-based` functions, particularly when dealing with `metadata` registration.

To solve this, the `wagmi` library is our recommended tool. It provides user-friendly hooks for Ethereum wallet interactions, ensuring proper account handling. [MetaMask](https://docs.metamask.io/wallet/how-to/connect/) also recommends the use of  `wagmi`.

You can employ wagmi to properly interface with the SDK as demonstrated below:
```
onst { data: wallet } = useWalletClient();

// Instantiate a new client for the SDK
const client = StoryClient.newClient({
  chainId: "sepolia",
  account: wallet!.account,
  transport: http(RPC),
});

// Utilize the signTypedData method on the account object
wallet?.account.signTypedData(/* ... */);
```
Remember that wagmi Core acts as a seamless interface on top of `viem`, facilitating wallet management and related operations.

I trust this clarifies the required approach and aids in the seamless integration of the account with the SDK. Should you need additional assistance or encounter more issues, please don't hesitate to get in touch.

Warm regards,

Bonnie

jacob-tucker:
Hi @bonnie57 , can you clarify what the purpose of `signTypedData` is? Where do I call that function?

I have setup the `wagmi` library in my DApp and configured the `StoryClient` like so: 
```
const config: StoryConfig = {
      account: wallet?.account as Account,
      transport: custom(wallet.transport),
[...]

PR #208, opened 2024-06-03
Export signature method

bonnie57:
## Description
- Fix export type issue
- Export signature method
- Modify calling signTypedData by wallet client, fix issue [Error when adding metadata in .register function with metamask](https://github.com/storyprotocol/sdk/issues/204)

## Test Plan 
image
image

### DIFF
--- a/packages/core-sdk/src/resources/nftClient.ts
+++ b/packages/core-sdk/src/resources/nftClient.ts
@@ -32,10 +32,9 @@ export class NftClient {
    * @returns A Promise that resolves to a CreateNFTCollectionResponse containing the transaction hash and collection address.
    * @emits CollectionCreated (nftContract);
    */
-  public async createNFTCollection<
-    TReq extends CreateNFTCollectionRequest,
-    TRes = CreateNFTCollectionResponse<TReq>,
-  >(request: TReq): Promise<TRes> {
+  public async createNFTCollection(
+    request: CreateNFTCollectionRequest,
+  ): Promise<CreateNFTCollectionResponse> {
     try {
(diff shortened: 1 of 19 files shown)


packages/core-sdk/src/resources/nftClient.ts
```diff
@@ -32,10 +32,9 @@ export class NftClient {
    * @returns A Promise that resolves to a CreateNFTCollectionResponse containing the transaction hash and collection address.
    * @emits CollectionCreated (nftContract);
    */
-  public async createNFTCollection<
-    TReq extends CreateNFTCollectionRequest,
-    TRes = CreateNFTCollectionResponse<TReq>,
-  >(request: TReq): Promise<TRes> {
```
bonnie57, line 38: Remove generic types to enhance SDK generation.

packages/core-sdk/src/types/common.ts
```diff
@@ -1,4 +1,18 @@
+import { Address, WalletClient } from "viem";
+
+import { SetPermissionsRequest } from "./resources/permission";
+
 export type TypedData = {
   interface: string; // i.e. "(address,uint256)"
   data: unknown[];
 };
+
+export type SignatureHelpParameter = {
```
edisonz0718, line 10: How is this parameter used? The naming is a bit weird. Can it just be SignatureParameter?

edisonz0718: I think it should just be getPermissionSignatureRequest

edisonz0718 (approved):

PR #265, opened 2024-09-25
Refactor client initialization logic to use a generic lazy-loaded method

cypherpepe:
Replaced individual client initializations with a getClient method that lazily loads and stores clients in a single object. This reduces code duplication and simplifies the management of client instances.

## Description

This PR refactors the client initialization logic by introducing a getClient method that lazily loads and caches clients in a single object. The previous approach initialized each client separately, leading to duplicated code and less maintainability. The new method reduces redundancy and makes the code more modular and easier to extend.

## Test Plan 

1. Ensure that lazy initialization occurs without errors and that subsequent calls retrieve the cached instance.
2. Confirm that client functionality is not impacted by the refactoring using both wallet and account configurations.

## Related Issue

No related issue. This is an internal refactoring to improve code structure.

## Notes

No additional front-end or navigation changes required.

### DIFF
--- a/packages/core-sdk/src/client.ts
+++ b/packages/core-sdk/src/client.ts
@@ -27,13 +27,7 @@ export class StoryClient {
   private readonly config: StoryConfig & { chainId: SupportedChainIds };
   private readonly rpcClient: PublicClient;
   private readonly wallet: SimpleWalletClient;
-  private _ipAsset: IPAssetClient | null = null;
-  private _permission: PermissionClient | null = null;
-  private _license: LicenseClient | null = null;
-  private _dispute: DisputeClient | null = null;
-  private _ipAccount: IPAccountClient | null = null;
-  private _royalty: RoyaltyClient | null = null;
-  private _nftClient: NftClient | null = null;
+  private _clients: { [key: string]: any } = {};
 
   /**
    * @param config - the configuration for the SDK client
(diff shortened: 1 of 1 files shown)


bonnie57 (commented):
Hi @cypherpepe,

The idea of a generic lazy-loaded method is excellent! I’ve left a few comments for potential enhancements. Please take a look when you have a chance.

packages/core-sdk/src/client.ts
```diff
@@ -27,13 +27,7 @@ export class StoryClient {
...
   private readonly rpcClient: PublicClient;
   private readonly wallet: SimpleWalletClient;
-  private _ipAsset: IPAssetClient | null = null;
-  private _permission: PermissionClient | null = null;
-  private _license: LicenseClient | null = null;
-  private _dispute: DisputeClient | null = null;
-  private _ipAccount: IPAccountClient | null = null;
-  private _royalty: RoyaltyClient | null = null;
-  private _nftClient: NftClient | null = null;
+  private _clients: { [key: string]: any } = {};
```
bonnie57, line 30: We need to create a type-safe client to avoid using `any`. For example, define client-specific types:

```typescript
type ClientRegistry = {
  ipAsset: IPAssetClient;
  permission: PermissionClient;
  // ...
};

File: packages/core-sdk/src/client.ts [hunk 2]
```diff
@@ -106,100 +100,45 @@ export class StoryClient {
   }
 
   /**
-   * Getter for the ip asset client. The client is lazily created when
+   * Generic getter for clients. The client is lazily created when
    * this method is called.
    *
-   * @returns the IPAssetClient instance
+   * @param clientName - the name of the client
+   * @param ClientClass - the class of the client
```
bonnie57, line 107: We now prefer using inline annotation. Could you use it as well?

packages/core-sdk/src/client.ts
```diff
@@ -106,100 +100,45 @@ export class StoryClient {
...
+   * @returns the client instance
    */
-  public get ipAsset(): IPAssetClient {
-    if (this._ipAsset === null) {
-      this._ipAsset = new IPAssetClient(this.rpcClient, this.wallet, this.config.chainId);
+  private getClient<T>(clientName: string, ClientClass: new (...args: any[]) => T): T {
```
bonnie57, line 110: If the client type is `ClientRegistry`, you can use the following code. You also need to consider the `chainId` inconsistency issue.

```typescript
private getClient<K extends keyof ClientRegistry>(
    clientName: K,
    ClientClass: new (
        rpcClient: PublicClient, 
        wallet: SimpleWalletClient, 
        chainId?: ChainIds
    ) => ClientRegistry[K] 
): ClientRegistry[K]

PR #318, opened 2024-11-18
Update index.ts to expose SimpleWalletClient object

roschler:
The SimpleWalletClient object is needed by developers creating Web3 apps that interact with the user's wallet.  For example:

        const config: StoryConfig = {
            chainId: await this.getCurrentChainId(), // Get the current chain ID from MetaMask
            transport: http(RPCProviderUrl), // Using MetaMask's provider
            wallet: this.walletClient as SimpleWalletClient
        };

This is so they can interact with the user's wallet for doing important operations on the Story Protocol blockchain, like signing payloads or paying for transactions.

## Description

Example:

- 1. add user login page.
- 2. ...

## Test Plan 

Example: 
- 2. ...

## Related Issue

Example: Issue #123

## Notes

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @DonFungible @edisonz0718 @jacob-tucker @AndyBoWu
+* @LeoHChen @DonFungible @edisonz0718 @jacob-tucker @AndyBoWu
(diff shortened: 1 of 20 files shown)


bonnie57:
I will close the pr due to a long time with no response.

roschler:
Hi @bonnie57.  For some reason I never saw an E-mail notification regarding your message asking me to create an issue.  However, I believe the change I suggested is no longer necessary.

PR #361, opened 2024-12-10
SDK Workflows Update

bpolania:
## Description
This github workflow upgrade updates the `pr-internal` workflow so it runs integration tests only when merging to `dev`.

* Based on the type of event, `pull_request` or `push`, it will run either unit test or integration tests reusable workflows.
* As a part of this upgrade two reusable workflows were created, here is the PR for that update: https://github.com/storyprotocol/gha-workflows/pull/100
* The Turbo script was also update to allow integration tests runs
* Removed the `timestamp` requirement for `push` events.

## Test Plan 
To test this completely it's necessary to make a PR and merge it.


package.json
```diff
@@ -10,6 +10,7 @@
     "lint": "turbo run lint",
     "fix": "turbo run fix",
     "test": "turbo run test --no-cache --concurrency=1",
+    "test:integration": "turbo run test:integration --no-cache --concurrency=1",
```
bonnie57, line 13: why do you limit the concurrency of 1? If you don't have a limit, you can run fast.

bpolania: Fixed

PR #382, opened 2025-01-12
Update permission.ts

offensif:
## Description
Fix typeScript type mismatch. In other parts of the code, the transaction hash is usually represented as a Hex type, not a string.

- 1. add user login page.
- 2. ...

## Test Plan 

Example: 
- 2. ...

## Notes
If the type remains as string, discrepancies and potential errors will arise when handling txHash in other parts of the application that expect a Hex value.


bonnie57 (commented):
Hi @offensif,

The `txHash` type looks good. Could you update this PR to address this issue across the entire repository? Additionally, it would be helpful if you could modify the PR title and description to better reflect the changes.

Issue #381, opened 2025-01-08
Add new methods in v1.3


bonnie57:
[priphery change.pdf](https://uploads.linear.app/56d6ac3c-6d4d-48b3-899a-c744b912f5b4/d8b715fe-5970-4c71-b0a3-f47662be0a59/1913c3b9-8188-47d1-9d2d-a267d8c5ab88?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzU2ZDZhYzNjLTZkNGQtNDhiMy04OTlhLWM3NDRiOTEyZjViNC9kOGI3MTVmZS01OTcwLTRjNzEtYjBhMy1mNDc2NjJiZTBhNTkvMTkxM2MzYjktODE4OC00N2QxLTlkMmQtYTI2N2Q4YzVhYjg4IiwiaWF0IjoxNzM2MzI1OTczLCJleHAiOjMzMzA2ODg1OTczfQ.7A4B2zSfV30Aer338p05JE1phR3QH-lmq3CfxZ8dmqw)

jason-halliday:
I'm curious about when SDK v1.3, which supports the V1.3 contract, is planned to be released.

PR #383, opened 2025-01-17
Update the related methods

bonnie57:
## Description
- Update the workflow methods
- Extract the `validateLicenseTermsData`,  `validateDerivativeData`, and `getLicenseData` methods
- Fix ABI conflict issue and remove override methods from abi
- Fix the related tests including unit tests and integration tests

## Test Plan 
image

## Related Issue
Add new methods in v1.3 #381


packages/core-sdk/src/abi/generated.ts
```diff
@@ -20165,44 +18442,6 @@ export class LicenseAttachmentWorkflowsClient {
     };
   }
 
-  /**
-   * method mintAndRegisterIpAndAttachPILTerms for contract LicenseAttachmentWorkflows
-   *
-   * @param request LicenseAttachmentWorkflowsMintAndRegisterIpAndAttachPilTerms3Request
```
DracoLi, line 20171: how did these get removed? i thought wagmi will generate all method in the contract and these are deprecated but still there.

bonnie57: I have communicated with Seb, and we don't need compatibility back, so Smart Contract has removed it.

jacob-tucker:
Hi @bonnie57 , I have one comment: Any time we have `maxRevenueShare`, we should expect user to pass in a value 0-100, not 0-100,000,000. Note that we already do this for some `commercialRevShare` fields like we do [here](https://github.com/storyprotocol/sdk/blob/dev/packages/core-sdk/src/utils/licenseTermsHelper.ts#L74).

packages/core-sdk/test/unit/resources/ipAsset.test.ts
```diff
@@ -43,6 +45,24 @@ const licenseTerms: LicenseTerms = {
   currency: MockERC20.address,
   uri: "",
 };
+
+const licensingConfig: LicensingConfig = {
```
DracoLi, line 49: no need to do in this PR, but I think we can move these mock data for unit tests into a constants file so its easier to track and reuse.

DracoLi (approved):

bonnie57:
> Hi @bonnie57 , I have one comment: Any time we have `maxRevenueShare`, we should expect user to pass in a value 0-100, not 0-100,000,000. Note that we already do this for some `commercialRevShare` fields like we do [here](https://github.com/storyprotocol/sdk/blob/dev/packages/core-sdk/src/utils/licenseTermsHelper.ts#L74).

I agree. ' maxRevenueShare is your expected behavior now. Did you see any different code from `commercialRevShare`?

packages/core-sdk/test/unit/resources/ipAsset.test.ts
bonnie57: I agree, but the object only works in the `ipAsset`.

DracoLi: one thing to note - this test file is getting large, same for the ipAsset resource. might be worth looking into if we can separate some of the logic in the file.

Issue #385, opened 2025-01-20
Redesign: license registerPILTerms


bonnie57:

DracoLi:
@bonnie57 I'm taking a look at this

PR #389, opened 2025-01-23
[WIP] Redesign with registerPILTerms methods

bonnie57:
## Description

Example:

- 1. add user login dpage.
- 2. ...

## Test Plan 

Example: 
- 2. ...

## Related Issue
Redesign: license registerPILTerms #385

## Notes


packages/core-sdk/src/types/resources/license.ts
```diff
@@ -4,92 +4,88 @@ import { TxOptions } from "../options";
...
+  /** The URI of the license terms, which can be used to fetch the offchain license terms.*/
+  uri: string;
+}
+/**
+ * The request object that contains all data needed to register a license term.
+ */
+export type RegisterPILTermsRequest<PILType extends PIL_TYPE> = {
+  /**  The license terms to be registered. */
+  terms:
+    | (PILType extends PIL_TYPE.COMMERCIAL_USE ? CommercialLicenseTerms : never)
```
DracoLi, line 57: I suggest we simplify this usage. For `terms` here in the request, the type should be `PILTermsInput`, which is the same as `RegisterPILTermsRequest` from before. `PilFlavor.xxx` returns `PILTerms`, which should be compatible with `PILTermsInput` as `PILTermsInput` is just `PILTerms` but with some attributes that is now required.
I don't need to allow the user to provide an enum to use a preset, but just need to do `PILFlavor.commercialUse({...})`

PR #390, opened 2025-01-23
Compatible sdk for devnet

bonnie57:
## Description
- Fix integration test including `dispute` and other tests
- Add `maxAllowedRewardShare` fields
- Add limitations about `defaultMintingFee`, `royaltyPolicy`, and `MintingFee`.
- Rename `devent ` to `homer`

## Test Plan 
image
image

## Related Issue
Compatible sdk for devnet #386

## Notes


packages/core-sdk/src/utils/chain.ts
```diff
@@ -1,18 +1,18 @@
 import { defineChain } from "viem/utils";
 
-export const odyssey = defineChain({
-  id: 15_16,
-  name: "odyssey",
+export const devnet = defineChain({
+  id: 13_15,
```
DracoLi, line 4: why not just `1315` here? is there a reason?

packages/core-sdk/test/unit/utils/utils.test.ts
```diff
@@ -152,13 +152,13 @@ describe("Test chainStringToViemChain", () => {
...
   });
 
-  it("should return iliad testnet if id is 1516", () => {
-    const chain = chainStringToViemChain("1516");
-    expect(chain).to.equal(odyssey);
+  it("should return iliad testnet if id is 1315", () => {
+    const chain = chainStringToViemChain("1315");
+    expect(chain).to.equal(devnet);
   });
   it("should return iliad testnet if id is iliad", () => {
```
DracoLi, line 159: these test name should be updated

DracoLi (approved):

packages/core-sdk/src/utils/chain.ts
bonnie57: In `viem`, it also follows the pattern, such as `sepolia`. It will become more readable.

PR #407, opened 2025-02-04
feat: added auto wip conversion for ipAssets

### LINKED ISSUE #403: Auto IP -> WIP conversation when mint or/and registering IPs
When minting and/or registering IP assets, the SDK should auto handle wrapping IP to WIP if the wallet does not have enough WIP to cover the fees. This behaviour should be the default but can be overridden via `wipOptions`

DracoLi:
## Description

Resolves #403

### DIFF
--- a/packages/core-sdk/src/utils/utils.ts
+++ b/packages/core-sdk/src/utils/utils.ts
@@ -93,12 +93,16 @@ export const chain: { [key in SupportedChainIds]: "1315" } = {
   1315: "1315",
 };
 
-export const validateAddress = (address: string): Address => {
+export function validateAddress(address: string): Address {
   if (!isAddress(address, { strict: false })) {
     throw Error(`Invalid address: ${address}`);
   }
   return address;
-};
+}
+
+export function validateAddresses(addresses: string[]): Address[] {
+  return addresses.map((address) => validateAddress(address));
+}
 
 /** @deprecated use {@link validateAddress} */
 export const getAddress = (address: string, name: string, chainId?: number): Address => {
@@ -110,6 +114,6 @@ export const getAddress = (address: string, name: string, chainId?: number): Add
   return checksumAddress(address, chainId);
 };
 
-export function getIPAmountDisplay(amount: bigint): string {
-  return `${formatEther(amount)}IP`;
+export function getTokenAmountDisplay(amount: bigint, unit = "IP"): string {
+  return `${formatEther(amount)}${unit}`;
 }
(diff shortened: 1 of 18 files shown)


packages/core-sdk/src/utils/utils.ts
```diff
@@ -93,12 +93,16 @@ export const chain: { [key in SupportedChainIds]: "1315" } = {
   1315: "1315",
 };
 
-export const validateAddress = (address: string): Address => {
+export function validateAddress(address: string): Address {
```
bonnie57, line 96: Why create validateAddress instead of getAddress methods? And I notice you don't use `checksumAddress'.

packages/core-sdk/src/utils/wipFeeUtils.ts
```diff
@@ -0,0 +1,268 @@
...
+      }
+      const encodedData = client.approveEncode(approval);
+      allCalls.push({
+        target: encodedData.to,
+        allowFailure: false,
+        value: 0n,
+        callData: encodedData.data,
+      });
+    });
+    return allCalls;
```
bonnie57, line 59: In the method, you don't implement `multicall to batch all approve calls'. You just throw call in next methods to multicall. I prefer multicall here, it will make this method make more sense, what do you think?

packages/core-sdk/src/utils/wipFeeUtils.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+  return [];
+};
+
+export const calculateLicenseWipMintFee = async (params: CalculateDerivativeMintFeeParams) => {
+  const fee = await params.licensingModuleClient.predictMintingLicenseFee({
+    licensorIpId: params.parentIpId,
+    licenseTemplate: params.licenseTemplateClient.address,
+    licenseTermsId: params.licenseTermsId,
+    amount: 1n,
```
bonnie57, line 78: I prefer to support dynamic amount instead of fixed number in these methods, it will be more flexible.

packages/core-sdk/src/utils/wipFeeUtils.ts
DracoLi: we canot do multicall here, it needs to collect all other calls, which is not handled by this method.

packages/core-sdk/src/utils/wipFeeUtils.ts
DracoLi: yeah i agree, i'll update this.

packages/core-sdk/src/utils/wipFeeUtils.ts
bonnie57: My concern is that the method is utility methods and should complete the approval of all spender features. However, for multicall, it's an intermediate step, which may cause confusion. I also cannot independently use the multicall feature with these methods.

PR #422, opened 2025-02-13
Dispute Integration tests fixed and expanded

bpolania:
### This PR addresses the following issues with the Dispute Integration tests:

* Some test were skipped
* Test cases were added so it's now possible to resolve disputes in `aeneid`

### The Following changes were implemented:

* `Utils` module were updated, so now it's possible to create StoryClients with a PK as a parameter
* In order to be able to resolve disputes the `setDisputeJudgement` method of the `DisputeModule.sol` contract is called by a whitelisted judging address. Therefore the `JUDGE_PRIVATE_KEY` variable was added to the `.env` file.
* CIDs can now be automatically generated with the `generateCID` method.

### About Judging Disputes

In order to resolved disputes, once a dispute has been created, it should be judged. This process will be processed by UMA on mainnet but it can be tested by impersonating a judge. This is the process implemented in this PR as follows:

* A whitelisted account is setup, this account can judge disputes.
* A user raises a dispute creating a `RaiseDisputeRequest` and calling the `raiseDispute` method. 
* A public client using the judge account calls  the `setDisputeJudgement` method of the `DisputeModule.sol` contract.
* The dispute is resolved. 

### Comments

* Before merging, the new environmental variables need to be set on github


packages/core-sdk/.env.example
```diff
@@ -1,2 +1,4 @@
 TEST_WALLET_ADDRESS = 0x0000000000000000000000000000000000000000
-WALLET_PRIVATE_KEY = 0x0000000000000000000000000000000000000000
\ No newline at end of file
+WALLET_PRIVATE_KEY = 0x0000000000000000000000000000000000000000
```
bonnie57, line 2: The account also needs to add workflow, otherwise, integration tests in the workflow will fail.

packages/core-sdk/package.json
```diff
@@ -52,7 +52,9 @@
     "c8": "^8.0.1",
     "chai": "^4.3.10",
     "chai-as-promised": "^7.1.1",
-    "eslint": "^8.50.0",
+    "eslint": "^8.57.1",
+    "eslint-config-prettier": "^8.10.0",
```
bonnie57, line 56: Is there another config for eslint?

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -1,22 +1,59 @@
...
+import * as sha256 from "multiformats/hashes/sha2";
+
 const expect = chai.expect;
 chai.use(chaiAsPromised);
 
+// Add the dispute module address and ABI
+const DISPUTE_MODULE_ADDRESS = "0x9b7A9c70AFF961C799110954fc06F3093aeb94C5";
+const SET_DISPUTE_JUDGEMENT_ABI = parseAbiItem(
+  "function setDisputeJudgement(uint256 disputeId, bool decision, bytes calldata data)",
+);
```
bonnie57, line 30: In my opinion, you need to get these data from `generated.ts`. Once smart contracts change, they will automatically change it, which is good to maintain.

packages/core-sdk/test/integration/utils/util.ts
```diff
@@ -138,3 +129,13 @@ export const getStoryClient = (): StoryClient => {
   };
   return StoryClient.newClient(config);
 };
+
+export const getStoryClientWithKey = (privateKey: `0x${string}`): StoryClient => {
```
bonnie57, line 133: You can reuse `getStoryClient` and permit it to accept an optional private key.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -1,22 +1,59 @@
...
+
+const judgeWalletClient = createWalletClient({
+  chain: chainStringToViemChain("aeneid"),
+  transport: http(RPC),
+  account: privateKeyToAccount(process.env.JUDGE_PRIVATE_KEY as `0x${string}`),
+});
```
bonnie57, line 36: It's the same as `getStoryClientWithKey`.

packages/core-sdk/.env.example
```diff
@@ -1,2 +1,4 @@
...
+JUDGE_PRIVATE_KEY = 0x0000000000000000000000000000000000000000
```
DracoLi, line 3: i think we should think about other ways to handle this. Since it seems this key cannot just be any private key that people can create. This will make it hard for someone to set this up and verify tests locally. One option is separate the tests that require this as its own integration tests that should only be run by CI and not meant to be run like the other integration tests locally.

packages/core-sdk/package.json
DracoLi: please include some info on why these are added/updated in the pr description.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -99,23 +136,74 @@ describe("Dispute Functions", () => {
...
+        liveness: 2592000,
+        bond: 0,
+        txOptions: {
+          waitForTransaction: true,
+        },
+      };
+      const response = await clientA.dispute.raiseDispute(raiseDisputeRequest);
+      disputeId = response.disputeId!;
+
+      // Setting dispute judgement directly through contract
```
DracoLi, line 173: can we add some details on how this works in the PR description? I think will help reviewer and others looking at the code to understand how we are testing disputes since its not very obvious.

packages/core-sdk/.env.example
```diff
@@ -1,2 +1,4 @@
...
+WALLET_PRIVATE_KEY_2 = 0x0000000000000000000000000000000000000000
```
DracoLi, line 4: Can we try to limit the amount of setup up work needed? Instead of relying on requiring two funded test wallets, can we see if we can create the second wallet programmatically and then have wallet 1 transfer enough tokens for gas?

packages/core-sdk/.env.example
bpolania: Yes. It will be added before merging

packages/core-sdk/test/integration/utils/util.ts
bpolania: Yeah, I thought about doing a single function, but `getStoryClient` is so widely used that I didn't want to break the whole SDK. I will merge them into a `getStoryClient`.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: done.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: See the other comment. Now there's a single `getStoryClient` method and `getStoryClientWithKey` was removed

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Added. 

```

### About Judging Disputes

In order to resolved disputes, once a dispute has been created, it should be judged. This process will be processed by UMA on mainnet but it can be tested by impersonating a judge. This is the process implemented in this PR as follows:

* A whitelisted account is setup, this account can judge disputes.
* A user raises a dispute creating a `RaiseDisputeRequest` and calling the `raiseDispute` method. 
* A public client using the judge account calls  the `setDisputeJudgement` method of the `DisputeModule.sol` contract.
* The dispute is resolved. 
```

packages/core-sdk/package.json
bpolania: Reverted back to previous configuration

packages/core-sdk/.env.example
bpolania: Done. It will check for the `JUDGE_PRIVATE_KEY` in the workflow, if it doesn't find it, it will skip run the tests.

packages/core-sdk/.env.example
bpolania: I was trying to avoid this because the integration tests are VERY slow, but I did the changes, it's now creating an funding a wallet each time the tests are ran.

bonnie57 (approved):

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -99,23 +141,87 @@ describe("Dispute Functions", () => {
...
         txOptions: { waitForTransaction: true },
       }),
     ).to.be.rejected;
   });
+
+  describe("resolveDispute", () => {
+    let disputeId: bigint;
+
+    before(function (this: Mocha.Context) {
+      if (!process.env.JUDGE_PRIVATE_KEY) {
```
DracoLi, line 166: :+1: great solution! I suggest we can also add some comments here on this key and why its skipped. Otherwise its not clear to other future devs reading this.

DracoLi (commented):
the pnpm-lock file shouldn't need to be changed now that there are no package changes

packages/core-sdk/test/integration/utils/util.ts
```diff
@@ -130,11 +121,12 @@ export const approveForLicenseToken = async (address: Address, tokenId: bigint)
   const hash = await walletClient.writeContract(call);
   await waitTx(publicClient, hash);
 };
-export const getStoryClient = (): StoryClient => {
+export const getStoryClient = (privateKey?: `0x${string}`): StoryClient => {
```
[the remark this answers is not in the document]
bpolania: Done

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Comments added

PR #423, opened 2025-02-13
Enhance contractCallWithFees with erc20

bonnie57:
### 1. New Dispute Method: `tagIfRelatedIpInfringed`
- Accept array to handle multiple IP ID
- Accept useMulticallWhenPossible option
- Add `tagIfRelatedIpInfringed` to dispute with muticall support.

### 4. Update Util methods: `contractCallWithWipFees` 

- Rename  `contractCallWithWipFees` to `contractCallWithFees` 

- `contractCallWithFees`  accept both ERC20 and Wrapper Ip

### 5. Other Changes
- Export types related to the WIP client, including `ERC20Options` and `TransactionResponse`.
- Add detailed documentation about `bond` and `liveness` in the `raiseDispute` method of the dispute module.
- Update the block creation fields for both the mainnet and Aeneid chain using data from the blockchain explorer.
- Add unit tests for the Wrapper Client and other components as needed.
## Test Plan 
Integration test
image


bpolania:
* It created a conflict with the new Dispute test cases, it should be an easy fix
* I discussed with @DracoLi that we will start migrating to a more standard comments system, after all this is a user facing code. In any case I wonder if we should start creating JSDoc compliant comments and docs. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

packages/core-sdk/src/utils/Erc20FeeUtils.ts
```diff
@@ -175,98 +173,108 @@ const multiCallWrapIp = async ({
...
-  if (startingBalance < totalFees) {
-    throw new Error(
-      `Wallet does not have enough IP to wrap to WIP and pay for fees. Total fees: ${getTokenAmountDisplay(
-        totalFees,
-      )}, balance: ${getTokenAmountDisplay(startingBalance)}`,
-    );
-  }
-  // error if there's enough IP to cover fees and we cannot wrap IP to WIP
-  if (!autoWrapIp) {
+  if (isWip) {
```
DracoLi, line 236: prefer return early and avoid large nesting if possible.
```
if (!isWip) {
  throw ...
}

// handle wip logic

File: packages/core-sdk/src/utils/chain.ts [hunk 2]
```diff
@@ -42,7 +42,7 @@ export const mainnet = defineChain({
   contracts: {
     multicall3: {
       address: "0xca11bde05977b3631167028862be2a173976ca11",
-      blockCreated: 5882,
+      blockCreated: 340998,
```
DracoLi, line 45: do you know what does this? what's the consequence of this not being the correct block before?

packages/wagmi-generator/wagmi.config.ts
```diff
@@ -287,7 +289,7 @@ export default defineConfig(async () => {
           ],
           RoyaltyPolicyLAP: ["onRoyaltyPayment", "getRoyaltyData"],
           LicenseToken: ["ownerOf"],
-          SPG: ["CollectionCreated"],
+          SPG: ["CollectionCreated", "mintFeeToken", "mintFee"],
```
DracoLi, line 292: adding these will be a separate PR?

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -1,75 +1,149 @@
...
-    const response = await clientA.dispute.raiseDispute(raiseDisputeRequest);
+    const response = await client.dispute.raiseDispute(raiseDisputeRequest);
+    disputeId = response.disputeId!;
     expect(response.txHash).to.be.a("string").and.not.empty;
     expect(response.disputeId).to.be.a("bigint");
   });
 
-  it("should throw error when liveness is out of bounds", async () => {
-    const minLiveness = await clientA.dispute.arbitrationPolicyUmaReadOnlyClient.minLiveness();
+  it.skip("should throw error when liveness is out of bounds", async () => {
```
DracoLi, line 124: note that a rebase from dev should fix these now i think.

packages/core-sdk/src/utils/token.ts
```diff
@@ -0,0 +1,72 @@
+import { Address, Hash, PublicClient } from "viem";
+
+import { EncodedTxData, Erc20Client, SimpleWalletClient, WrappedIpClient } from "../abi/generated";
+
+export interface TokenClient {
```
DracoLi, line 5: if this is mainly to allow us to pass a single erc20 client to the multicall function, I feel like we can just use `Erc20Client` for both, then a separate wipClient if the user needs to deposit wip instead of creating a wrapper for both. WIP and ERC20 have the same abi to do balance and approval checks.

DracoLi (commented):
lots of great changes. Next time should consider splitting this into separate PRs for quicker review.

bonnie57:
> * It created a conflict with the new Dispute test cases, it should be an easy fix
> * I discussed with @DracoLi that we will start migrating to a more standard comments system, after all this is a user facing code. In any case I wonder if we should start creating JSDoc compliant comments and docs. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

Hi Boris, the conflict has been solved. And I do some enhance, you can review it.

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -8,45 +8,40 @@ import {
...
-   *   @param request.bond The bond size.
-   *   @param request.txOptions [Optional] This extends `WaitForTransactionReceiptParameters` from the Viem library, excluding the `hash` property.
-   * @returns A Promise that resolves to a RaiseDisputeResponse containing the transaction hash.
-   * @throws `NotRegisteredIpId` if targetIpId is not registered in the IPA Registry.
-   * @throws `NotWhitelistedDisputeTag` if targetTag is not whitelisted.
-   * @throws `ZeroLinkToDisputeEvidence` if linkToDisputeEvidence is empty
-   * @calls raiseDispute(address _targetIpId, string memory _linkToDisputeEvidence, bytes32 _targetTag, bytes calldata _data) external nonReentrant returns (uint256) {
-   * @emits DisputeRaised (disputeId_, targetIpId, msg.sender, arbitrationPolicy, linkToDisputeEvidence, targetTag, calldata);
+   * Raises a dispute on a given ipId.
+   * Emits DisputeRaised (disputeId_, targetIpId, msg.sender, arbitrationPolicy, linkToDisputeEvidence, targetTag, calldata);
```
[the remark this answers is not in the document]
bonnie57: I agree that users should refer to the POC docs for the most updated errors and events emitted by the POC itself. However, I think including these POC tx events in our documentation could still be valuable for users who want to understand the details of how the SDK interacts with the POC. It provides additional context and helps users debug or trace issues more effectively.

That said, I understand the concern about keeping the documentation focused on the SDK's responsibilities. Perhaps we can include these events in a separate section or appendix, clearly labeled as "POC Transaction Events" with a note that these are not directly emitted by the SDK but are relevant for understanding the full flow. This way, we maintain a clear distinction while still providing useful information for advanced users.

What do you think about this approach?

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -185,4 +180,47 @@ export class DisputeClient {
...
+  ): Promise<TransactionResponse[]> {
+    try {
+      const objects: DisputeModuleTagIfRelatedIpInfringedRequest[] = request.args.map((arg) => ({
+        ipIdToTag: validateAddress(arg.ipIdToTag),
+        infringerDisputeId: BigInt(arg.infringerDisputeId),
+      }));
+      let txHashes: Hex[] = [];
+      if (
+        (request.useMulticallWhenPossible === undefined && request.args.length === 1) ||
+        request.useMulticallWhenPossible === false
```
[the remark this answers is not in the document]
bonnie57: I don't change anything. I remember we talk if user passes only a arg, we prefer directly call `tagIfRelatedIpInfringed`. So

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -47,7 +47,7 @@ export class RoyaltyClient {
   public ipRoyaltyVaultImplReadOnlyClient: IpRoyaltyVaultImplReadOnlyClient;
   public ipRoyaltyVaultImplEventClient: IpRoyaltyVaultImplEventClient;
   public multicall3Client: Multicall3Client;
-  public wipClient: WrappedIpClient;
+  public wrappedIpClient: WrappedIpClient;
```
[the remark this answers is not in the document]
bonnie57: When I read the code, I felt confused about which is internal `wrappedIp` instead of external `wip`. We need to distinguish it.
Besides, what do you mean about API breaking release?

packages/core-sdk/src/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: In this case, I'd rather we link to the POC docs on github instead of rewriting them here. We don't want 2 sources of truth.

packages/core-sdk/src/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: yep sorry, I think I miss read, req is not changed here. We should not use multicall if there's only one ip, so we should not support it.
This condition can be made clearer with `useMultiCall = request.useMulticallWhenPossible !== false && request.args.length > 1`, meaning only multicall if user did not set it to false and we have more than one ip.

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -39,3 +46,19 @@ export type ResolveDisputeResponse = {
   txHash?: string;
   encodedTxData?: EncodedTxData;
 };
+
+export type TagIfRelatedIpInfringedRequest = {
+  args: {
```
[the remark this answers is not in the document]
bonnie57: How about `infringementTags ` instead of `args`.  For ip, we habitually call ipId.

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -39,3 +46,19 @@ export type ResolveDisputeResponse = {
...
+    /** The dispute id that tagged the related infringing ipId */
+    infringerDisputeId: number | string | bigint;
+  }[];
+  /**
+   * Use multicall to batch the calls into one transaction when possible.
+   *
+   * @default true
+   * If args length is 1, multicall is not used.
+   */
+  useMulticallWhenPossible?: boolean;
```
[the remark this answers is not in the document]
bonnie57: For business requirements, not every method supports` multicall`. Including it in `txOption` could lead to confusion. Regarding `multicallOptions`, aside from this field, what other fields should we include? If there’s only one field, requiring users to pass it as an object doesn’t seem like a good practice. What are your thoughts?

packages/core-sdk/src/types/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: 👍

packages/core-sdk/src/resources/dispute.ts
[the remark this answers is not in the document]
bonnie57: Agree!

packages/core-sdk/src/utils/token.ts
bonnie57: Sync in the meeting.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -169,27 +187,100 @@ describe("Dispute Functions", () => {
...
       // Set up judge wallet client using whitelisted account
-      const judgeWalletClient = createWalletClient({
+      judgeWalletClient = createWalletClient({
         chain: chainStringToViemChain("aeneid"),
         transport: http(RPC),
-        account: privateKeyToAccount(process.env.JUDGE_PRIVATE_KEY as `0x${string}`),
+        account: privateKeyToAccount(process.env.JUDGE_PRIVATE_KEY as Address),
+      });
+
+      // Setup NFT collection
```
DracoLi, line 210: does this need to be in before each? Try to reuse and avoid submitting more tx that needed. This should be considered for all tx in a beforeEach block.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -169,27 +187,100 @@ describe("Dispute Functions", () => {
...
+          maxRts: 5 * 10 ** 6,
+          maxRevenueShare: 100,
+        },
+        allowDuplicates: true,
+        txOptions: { waitForTransaction: true },
+      });
+      childIpId = derivativeIpIdResponse.ipId!;
+    });
+
+    it("should raise a dispute", async () => {
```
DracoLi, line 281: does this test need to be here? We already have a test for raising a dispute, this is more of a setup step so don't need another test.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -210,9 +305,21 @@ describe("Dispute Functions", () => {
         args: [disputeId, true, "0x"],
         account: judgeWalletClient.account!,
       });
-
       const txHash = await judgeWalletClient.writeContract(request);
       await publicClient.waitForTransactionReceipt({ hash: txHash });
+
+      // Step 2: Tag derivative IP as infringing
+      const results = await clientA.dispute.tagIfRelatedIpInfringed({
```
DracoLi, line 312: i think also need a multicall test

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -210,9 +305,21 @@ describe("Dispute Functions", () => {
...
+        infringementTags: [
+          {
+            ipId: childIpId,
+            disputeId: disputeId,
+          },
+        ],
+        txOptions: { waitForTransaction: true },
+        options: { useMulticallWhenPossible: true },
```
DracoLi, line 320: default is true, no need to include this.

packages/core-sdk/test/unit/utils/FeeUtils.test.ts
```diff
@@ -3,26 +3,26 @@ import * as sinon from "sinon";
 import chaiAsPromised from "chai-as-promised";
```
DracoLi, line 3: lower case for filenames. `feeUtils.test.ts`

packages/wagmi-generator/wagmi.config.ts
```diff
@@ -185,8 +185,9 @@ export default defineConfig(async () => {
         [mainnetChainId]: "0xf96f2c30b41Cb6e0290de43C8528ae83d4f33F89",
       },
     },
+    // Make sure ERC20's abi is equal to the erc20 abi of viem
```
DracoLi, line 188: can remove

packages/core-sdk/test/unit/resources/royalty.test.ts
```diff
@@ -22,14 +27,23 @@ describe("Test RoyaltyClient", () => {
...
     walletMock.account = accountMock;
     royaltyClient = new RoyaltyClient(rpcMock, walletMock);
-    sinon.stub();
   });
 
   afterEach(() => {
     sinon.restore();
   });
 
   describe("Test royaltyClient.payRoyaltyOnBehalf", async () => {
```
DracoLi, line 36: i think we also need an unit test for payRoyaltyOnBehalf with erc20 token, making sure approve is called.

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -8,45 +8,41 @@ import {
...
-   *   @param request.txOptions [Optional] This extends `WaitForTransactionReceiptParameters` from the Viem library, excluding the `hash` property.
-   * @returns A Promise that resolves to a RaiseDisputeResponse containing the transaction hash.
-   * @throws `NotRegisteredIpId` if targetIpId is not registered in the IPA Registry.
-   * @throws `NotWhitelistedDisputeTag` if targetTag is not whitelisted.
-   * @throws `ZeroLinkToDisputeEvidence` if linkToDisputeEvidence is empty
-   * @calls raiseDispute(address _targetIpId, string memory _linkToDisputeEvidence, bytes32 _targetTag, bytes calldata _data) external nonReentrant returns (uint256) {
-   * @emits DisputeRaised (disputeId_, targetIpId, msg.sender, arbitrationPolicy, linkToDisputeEvidence, targetTag, calldata);
+   * Raises a dispute on a given ipId.
+   * Submits a {@link DisputeRaised} event.
+   * @see https://github.com/storyprotocol/protocol-core-v1/blob/main/contracts/interfaces/modules/dispute/IDisputeModule.sol
```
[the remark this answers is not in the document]
bonnie57: How about follow the below code:
```
   * @see {@link https://github.com/storyprotocol/protocol-core-v1/blob/main/contracts/interfaces/modules/dispute/IDisputeModule.sol | IDisputeModule.sol}

```
It's better to read. And `jsdoc` has the same example.
image

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -1,15 +1,26 @@
...
 
-import { TxOptions } from "../options";
+import { TxOptions, WithTxOptions } from "../options";
 import { EncodedTxData } from "../../abi/generated";
 
-export type RaiseDisputeRequest = {
+export type RaiseDisputeRequest = WithTxOptions & {
+  /** The IP ID that is the target of the dispute. */
   targetIpId: Address;
   cid: string;
```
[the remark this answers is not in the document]
bonnie57: Do you mean how to get the `targetIpId`?

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -1,15 +1,26 @@
...
+  /**
+   * The target tag of the dispute.
+   * @see https://docs.story.foundation/docs/dispute-module#dispute-tags
+   * @example "IMPROPER_REGISTRATION"
+   */
   targetTag: string;
+  /** The liveness is the time window in which a counter dispute can be presented (30days). The unit is second. */
   liveness: bigint | number | string;
+  /**
+   * The amount that the dispute initiator pays upfront into a pool.
```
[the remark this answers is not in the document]
bonnie57: The amount is ERC20. Do we need to add more details? I think it's clear.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: Yes, we need to make a derivative ipId.

packages/core-sdk/src/types/utils/wip.ts
```diff
@@ -67,27 +45,31 @@ export type WipApprovalCall = {
...
+  multicall3Address: Address;
+  /** all possible spenders of the erc20 token */
+  tokenSpenders: Erc20Spender[];
   contractCall: () => Promise<Hash>;
   encodedTxs: EncodedTxData[];
   rpcClient: PublicClient;
   wallet: SimpleWalletClient;
   sender: Address;
+  options: {
+    wipOptions?: WipOptions;
```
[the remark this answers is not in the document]
bonnie57: For example, the payRoyaltyOnBehalf method accepts a token argument, which can be either an ERC20 or WIP token. I have encapsulated the logic for handling WIP or ERC20 tokens within the method itself, so the caller does not need to worry about the underlying implementation details.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: It's a complete flow.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: Integration tests will not be in every case. Besides, `multicall` test has been covered by other cases with different options.

packages/wagmi-generator/wagmi.config.ts
bonnie57: It's the information to remind developers to keep the equal about standard ERC20. Such as wrapperIP is also the erc20, but it has a different abi.

packages/core-sdk/test/unit/resources/royalty.test.ts
bonnie57: I agree, I have already added it in line 86.

packages/core-sdk/src/resources/dispute.ts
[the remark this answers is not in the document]
bonnie57: I believe it's better to reference a specific event rather than providing a generic event link, as users may not know which event will be triggered. Additionally, since the contract is audited and unlikely to change frequently, I prefer the following format:
```
   * @see {@link https://github.com/storyprotocol/protocol-core-v1/blob/main/contracts/interfaces/modules/dispute/IDisputeModule.sol | DisputeRaised}

```
what do you think?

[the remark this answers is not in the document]
DracoLi: I think if the link is the name of the event, then the link should link to it instead too of the whole file.

packages/core-sdk/src/types/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: `cid`

packages/core-sdk/src/types/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: erc20 for what token address. its not clear to me thus I had to clarify.

packages/core-sdk/src/types/utils/wip.ts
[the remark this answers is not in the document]
DracoLi: yes you handled it in `payRoyaltyOnBehalf`, but in the future another dev might need to use this same util and thus will need to understand that even though you have 2 options here, only 1 can work at a time.

packages/core-sdk/test/integration/dispute.test.ts
DracoLi: i'm referring to the nft collection. You need a new collection every test?

packages/core-sdk/test/integration/dispute.test.ts
DracoLi: please explain why we have two tests called `should raise a dispute`, all testing the same thing. Its a complete flow does not explain why that needs to happen.

packages/core-sdk/test/integration/dispute.test.ts
DracoLi: which case. integration tests should cover critical path. is using `tagIfRelatedIpInfringed` with multiple tags not a critical path?

packages/wagmi-generator/wagmi.config.ts
DracoLi: this does not tell me this. i'm very confused what this comment is. I think if you want to convey that, it needs to be more informative.

packages/core-sdk/test/integration/dispute.test.ts
DracoLi: which existing case? Having multical tests with ipAsset does not mean any other multicall related functionalities does not need to be tested. Integration tests should cover critical path and I think using `tagIfRelatedIpInfringed` with multiple tags is a critical path as its a very common use case.

packages/core-sdk/src/types/utils/wip.ts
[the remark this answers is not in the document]
DracoLi: yes you handled it in `payRoyaltyOnBehalf`, but in the future another dev might need to use this same util function and thus will need to understand that even though you have 2 options here, only 1 can work at a time. Versus just making it impossible to provide two options.

packages/wagmi-generator/wagmi.config.ts
DracoLi: this does not tell me this, i thought its a reminder for yourself. I think if you want to convey that for future devs, it needs to be more informative.

packages/core-sdk/src/types/resources/dispute.ts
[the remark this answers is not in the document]
bonnie57: Sorry, I don't get the point. Which one needs to clarify?
The doc like:
  ```
 * The amount of ERC20 that the dispute initiator pays upfront into a pool.
 ```

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: No need, but I use SPG methods instead of nft contract. It's not compatible, so I need to create one.

DracoLi: nvm i see this is already not in a beforeEach. this is good then.

packages/core-sdk/src/types/utils/wip.ts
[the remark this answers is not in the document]
bonnie57: Yeah, I agree for a little confused. The option needs to be refactored. I think we don't distinguish `wipOption` or `ERC20Options`. Because they are both ERC20 tokens. You mentioned you will refactor in the 1.4, so for now I think it's good. It's the internal method.

And for the caller perspective, such as `payRoyaltyOnBehalf`, I don't care if the token is wip or erc20Token, I just pass to the `contractCallWithFees` method.  And I think accepting the erc20 token will become general, so we don't care in many places.  `contractCallWithFees` needs to handle the logic.

packages/core-sdk/test/integration/dispute.test.ts
DracoLi: should be moved to before. If you think it can cause confusing, please add a comment.

bonnie57: Have moved into before method. Please check it.

packages/core-sdk/src/types/resources/dispute.ts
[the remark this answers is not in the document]
bonnie57: I am not sure, let me confirm with the developer.

packages/core-sdk/src/resources/dispute.ts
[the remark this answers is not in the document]
DracoLi: also lets link it to a specific release instead of `main` code. we will just need to remember to update if new releases changes the logic.

packages/core-sdk/src/types/options.ts
```diff
@@ -1,16 +1,82 @@
...
+   *
+   * @default true
+   */
+  enableAutoApprove?: boolean;
+};
+/**
+ * Options to override the default behavior of the auto approve logic
+ */
+export type WithERC20Options = {
+  ERC20Options?: ERC20Options;
```
DracoLi, line 34: this need to be lower case, `erc20Options`

bonnie57: `ERC` is a distinctive term whose capitalization may vary in accordance with the customary practices of the team.

DracoLi (approved):
Looks good, good to merge 🚀

packages/core-sdk/src/types/options.ts
DracoLi: variable names cannot start capitalized. i'll add a lint if this cannot be enforced.

DracoLi: @bonnie57 please review and do additional research on some of my feedback and understand why I suggested them. Its clear in many instances that you need additional understanding of coding practices, conventions, and etc. I will not be able to teach all of this items through PR reviews.

DracoLi: General practices like provide clear documentations that other's can understand, use good variable names, do not mutate request objects, use camelCase for variable names, are all pretty common that I expect to not need a guideline for. 
Please ensure that similar issues that I've provided feedback on in this PR do not appear again in future PRs. Ensure you review your own code before putting them up for review.

PR #425, opened 2025-02-17
Add default value

bonnie57:
## Description
Add default value including :
### Derivative Data
-`maxMintingFee`: Default 0
- `maxRts`: 100_000_000
- `maxRevenueShare`: 100 (this is a percentage, SDK converts to 100m before calling protocol)

###  Licensing Config
- `isSet`: false
- `mintingFee`: 0
- `licensingHook`: zeroAddress
- `hookData`: zeroAddress
- `commercialRevShare`: 0
- `disabled`: false
- `expectMinimumGroupRewardShare`: 0
- `expectGroupRewardPool`: zeroAddress


packages/core-sdk/src/resources/group.ts
```diff
@@ -107,36 +107,8 @@ export class GroupClient {
...
-   *     @param request.ipMetadata.ipMetadataURI [Optional] The URI of the metadata for the IP.
-   *     @param request.ipMetadata.ipMetadataHash [Optional] The hash of the metadata for the IP.
-   *     @param request.ipMetadata.nftMetadataURI [Optional] The URI of the metadata for the NFT.
-   *     @param request.ipMetadata.nftMetadataHash [Optional] The hash of the metadata for the IP NFT.
-   *   @param request.txOptions [Optional] This extends `WaitForTransactionReceiptParameters` from the Viem library, excluding the `hash` property.
-   * @returns A Promise that resolves to a transaction hash, and if encodedTxDataOnly is true, includes encoded transaction data, and if waitForTransaction is true, includes IP ID, token ID.
-   * @emits IPRegistered (ipId, chainId, tokenContract, tokenId, resolverAddr, metadataProviderAddress, metadata)
+  /** Mint an NFT from a SPGNFT collection, register it with metadata as an IP, attach license terms to the registered IP, and add it to a group IP.
+   * it emits IPRegistered (ipId, chainId, tokenContract, tokenId, resolverAddr, metadataProviderAddress, metadata).
```
bonnie57, line 111: I changed it before we talked, so please keep it. Next doc change will create the separate pr. @DracoLi

DracoLi: all of these can be updated now with the format we talked about in https://github.com/storyprotocol/sdk/pull/423

packages/core-sdk/src/types/common.ts
```diff
@@ -9,25 +9,58 @@ export type TypedData = {
...
+  disabled?: boolean;
+  /**
+   * The minimum percentage of the group’s reward share (from 0 to 100%, represented as 100_000_000) that can be allocated to the IP when it is added to the group.
+   * If the remaining reward share in the group is less than the minimumGroupRewardShare,the IP cannot be added to the group.
+   * @default 0
+   */
+  expectMinimumGroupRewardShare?: number | string;
+  /**
+   * The address of the expected group reward pool. The IP can only be added to a group with this specified reward pool address, or address(0) if the IP does not want to be added to any group.
+   * @default address(0)
```
DracoLi, line 54: `address(0)` is not a valid value in javascript

packages/core-sdk/src/types/common.ts
```diff
@@ -9,25 +9,58 @@ export type TypedData = {
 };
 
 export type IpMetadataAndTxOptions = WithTxOptions & {
+  /** The desired metadata for the newly minted NFT and newly registered IP. */
   ipMetadata?: Partial<IpMetadataForWorkflow>;
 };
-
 export type LicensingConfig = {
```
DracoLi, line 15: this does not look right.
> Licensing Config: Licensing config property will be optional. If not provided, we will use the following defaults. If provided, user is required to fill ALL properties.

All properties are required, the config is not required.

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -8,10 +8,24 @@ import { IpMetadataForWorkflow } from "../../utils/getIpMetadataForWorkflow";
 
 export type DerivativeData = {
   parentIpIds: Address[];
+  /** The IDs of the license terms that the parent IP supports. */
   licenseTermsIds: bigint[] | string[] | number[];
+  /**
+   * The maximum minting fee that the caller is willing to pay. if set to 0 then no limit.
+   * @default 0
+   */
   maxMintingFee: bigint | string | number;
```
DracoLi, line 17: this is not optional?

packages/core-sdk/test/unit/resources/ipAsset.test.ts
```diff
@@ -924,7 +924,6 @@ describe("Test IpAssetClient", () => {
         ],
         allowDuplicates: false,
         recipient: "0x73fcb515cee99e4991465ef586cfe2b072ebb512",
```
DracoLi, line 926: need unit tests for the default value changes.

packages/core-sdk/src/types/common.ts
bonnie57: Who adds the doc?  Maybe we need to confirm with him about the real requirements.

DracoLi: i wrote it in the doc. if there was any confusion, then you need to bring it up before working on this.

DracoLi: i wrote this requirement weeks ago, it should be clear and I've received no questions on it. If it was unclear, then please bring it up **before** working on it. Its important to understand all requirements before starting.

Basically to help with migration, make licensing config optional. All properties in it should be required. For Derivative Data, all properties should be optional. Why? Because doing this will allow users using SDK 1.2.x to migrate with minimal code changes.

DracoLi: i wrote this requirement, as with all the other items in the doc. I think its clear and there was no questions regarding this item. If it was unclear, please bring it up before working on it. Its important to understand all requirements before starting.

Basically to help with migration, make licensing config optional. All properties in it should be required. For Derivative Data, the new properties should be optional. Why? Because doing this will allow users using SDK 1.2.x to migrate with minimal code changes as licensing config does not exist in 1.2.x.

bonnie57: Got it. When I work, I think it's clear. I'm confused by this sentence
> Make the following properties optional for all functions that requires it. SDK will set defaults if not provided. This is to help with SDK v1.2.x migration.

bonnie57: How about `setLicensingConfig` in the license module? Do we need to make `licenseConfig` optional?

DracoLi: No, the goal is to help with ip asset registration migration from `v1.2.x`, only those should be updated since those are not required before.

packages/core-sdk/src/types/resources/group.ts
```diff
@@ -2,27 +2,38 @@ import { Address } from "viem";
...
   groupId: Address;
+  /** Set to true to allow minting an NFT with a duplicate metadata hash. */
   allowDuplicates: boolean;
+  /** The maximum reward share percentage that can be allocated to each member IP. */
   maxAllowedRewardShare: number | string;
+  /** The data of the license and its configuration to be attached to the new group IP. */
   licenseData: LicenseData[];
+  /**
+   * The address of the recipient of the minted NFT.
+   * @default wallet address
```
DracoLi, line 30: `@default` should be an actual value. "wallet address" is not a valid address. We shouldn't need to add example for an address so lets remove.

packages/core-sdk/src/types/resources/group.ts
```diff
@@ -2,27 +2,38 @@ import { Address } from "viem";
...
+   */
   recipient?: Address;
+  /**
+   * The deadline for the signature in seconds.
+   * @default 1000s
```
DracoLi, line 35: this needs to be a valid value, so `1000`.

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -79,13 +106,13 @@ export type MintAndRegisterIpAssetWithPilTermsResponse = {
...
   nftContract: Address;
   tokenId: string | number | bigint;
+  /**
+   * The deadline for the signature in seconds.
+   * @default 1000s
+   */
   deadline?: string | number | bigint;
+  /** The derivative data to be used for register derivative. */
   derivData: DerivativeData;
-  sigMetadataAndRegister?: {
```
DracoLi, line 84: how come this is removed?

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -111,11 +143,16 @@ export type RegisterIpAndAttachPilTermsResponse = {
...
   tokenId?: bigint;
 };
-
 export type MintAndRegisterIpAndMakeDerivativeRequest = {
   spgNftContract: Address;
+  /** The derivative data to be used for register derivative. */
   derivData: DerivativeData;
+  /**
+   * Authors of the IP and their shares of the royalty tokens.
+   * @default wallet address
```
DracoLi, line 152: remove

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -111,11 +143,16 @@ export type RegisterIpAndAttachPilTermsResponse = {
...
+   */
   recipient?: Address;
+  /** Set to true to allow minting an NFT with a duplicate metadata hash. */
   allowDuplicates: boolean;
```
DracoLi, line 156: just realized we need to make this optional as well.

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -221,10 +258,13 @@ export type MintAndRegisterIpRequest = {
...
 } & IpMetadataAndTxOptions;
-
 export type RegisterPilTermsAndAttachRequest = {
   ipId: Address;
-  licenseTermsData: LicenseTermsData<RegisterPILTermsRequest, LicensingConfig>[];
+  /** The data of the license and its configuration to be attached to the IP. */
+  licenseTermsData: LicenseTermsData[];
+  /** The deadline for the signature in seconds.
+   * @default 1000s
```
DracoLi, line 266: update to valid value

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -321,7 +366,12 @@ export type DistributeRoyaltyTokens = {
   txOptions?: Omit<TxOptions, "encodedTxDataOnly">;
 };
 export type RoyaltyShare = {
+  /** The address of the recipient. */
   recipient: Address;
+  /**
+   *  The percentage of the royalty share.
+   *  @example 10 represents 10% which is 100_000_00.
```
DracoLi, line 373: `@example` should be for just the example value. You can do this. 
```suggestion
   *  The percentage of the total royalty share. For example, a value of 10 represents 10% of max royalty shares, which is 10,000,000
   * 
   *  @example 10
```

packages/core-sdk/src/types/resources/group.ts
bonnie57: If remove @default. we can add the annotation :
> If not provided, the function will use the user's own wallet address.

What do you think?

packages/core-sdk/src/types/resources/ipAsset.ts
bonnie57: We actually don't user the property in the code. It's the useless code.

packages/core-sdk/src/types/resources/group.ts
DracoLi: thats perfect

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -1,17 +1,31 @@
...
+   * @default 0
+   */
+  maxMintingFee?: bigint | string | number;
+  /**
+   *  The maximum number of royalty tokens that can be distributed to the external royalty policies (max: 100,000,000).
+   * @default 100_000_000
+   */
+  maxRts?: number | string;
+  /**
+   * The maximum revenue share percentage allowed for minting the License Tokens. Must be between 0 and 100 (where 100% represents 100_000_000).
```
DracoLi, line 24: :+1: very clear

DracoLi (approved):
🚀

PR #426, opened 2025-02-18
Add new methods for nftClient

bonnie57:
### Add Auto Wrapping Support to the following ipAsset functions
  - `mintAndRegisterIp`: Pay nft minting fee
  - `registerDerivative`
### Add following methods to `nftClient`. This is to allow dev to easily to get the mint fee of a SPG collection.
 - `mintFeeToken`: Gets the mint fee token of an spg nft contract
 - `mintFee`: Gets the mint fee of an spg nft contract

### DIFF
--- a/packages/core-sdk/src/resources/group.ts
+++ b/packages/core-sdk/src/resources/group.ts
@@ -44,6 +44,7 @@ import { getFunctionSignature } from "../utils/getFunctionSignature";
 import { validateLicenseConfig } from "../utils/validateLicenseConfig";
 import { getIpMetadataForWorkflow } from "../utils/getIpMetadataForWorkflow";
 import { getRevenueShare } from "../utils/licenseTermsHelper";
+import { RevShareType } from "../types/common";
 
 export class GroupClient {
   public groupingWorkflowsClient: GroupingWorkflowsClient;
@@ -147,7 +148,9 @@ export class GroupClient {
         spgNftContract: getAddress(spgNftContract, "request.spgNftContract"),
         recipient:
           (recipient && getAddress(recipient, "request.recipient")) || this.wallet.account!.address,
-        maxAllowedRewardShare: BigInt(getRevenueShare(request.maxAllowedRewardShare)),
+        maxAllowedRewardShare: BigInt(
+          getRevenueShare(request.maxAllowedRewardShare, RevShareType.MAX_ALLOWED_REWARD_SHARE),
+        ),
         licensesData: this.getLicenseData(request.licenseData),
         ipMetadata: getIpMetadataForWorkflow(request.ipMetadata),
         sigAddToGroup: {
@@ -257,7 +260,9 @@ export class GroupClient {
         licensesData: this.getLicenseData(request.licenseData),
         ipMetadata: getIpMetadataForWorkflow(request.ipMetadata),
         tokenId: BigInt(request.tokenId),
-        maxAllowedRewardShare: BigInt(getRevenueShare(request.maxAllowedRewardShare)),
+        maxAllowedRewardShare: BigInt(
+          getRevenueShare(request.maxAllowedRewardShare, RevShareType.MAX_ALLOWED_REWARD_SHARE),
+        ),
         sigAddToGroup: {
           signer: getAddress(this.wallet.account!.address, "wallet.account.address"),
           deadline: calculatedDeadline,
--- a/packages/core-sdk/src/resources/ipAsset.ts
+++ b/packages/core-sdk/src/resources/ipAsset.ts
@@ -105,7 +105,7 @@ import { AccessPermission } from "../types/resources/permission";
 import { LicenseTerms, RegisterPILTermsRequest } from "../types/resources/license";
 import { MAX_ROYALTY_TOKEN, royaltySharesTotalSupply } from "../constants/common";
 import { getFunctionSignature } from "../utils/getFunctionSignature";
-import { LicensingConfig } from "../types/common";
+import { LicensingConfig, RevShareType } from "../types/common";
 import { validateLicenseConfig } from "../utils/validateLicenseConfig";
 import { getIpMetadataForWorkflow } from "../utils/getIpMetadataForWorkflow";
 import {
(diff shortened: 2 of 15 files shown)


packages/core-sdk/test/integration/nftClient.test.ts
```diff
@@ -138,4 +140,14 @@ describe("nftClient Functions", () => {
       ).to.be.rejectedWith("Invalid mint fee token address");
     });
   });
+
+  it("should successfully get mint fee token", async () => {
+    const mintFeeToken = await client.nftClient.getMintFeeToken(spgNftContract);
+    expect(mintFeeToken).to.be.a("string").and.not.empty;
```
DracoLi, line 146: verify the values instead of just not empty.  same for mint fee

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -459,9 +461,13 @@ export type CommonRegistrationHandlerParams = WithWipOptions & {
   txOptions?: TxOptions;
 };
 
-export type CommonRegistrationResponse = {
+export type RegistrationResponse = {
   txHash?: Hex;
-  ipId?: Address;
-  tokenId?: bigint;
   receipt?: TransactionReceipt;
+  ipId?: Hex;
```
DracoLi, line 467: why is this changed to hex?

packages/core-sdk/src/resources/group.ts
```diff
@@ -147,7 +148,9 @@ export class GroupClient {
         spgNftContract: getAddress(spgNftContract, "request.spgNftContract"),
         recipient:
           (recipient && getAddress(recipient, "request.recipient")) || this.wallet.account!.address,
-        maxAllowedRewardShare: BigInt(getRevenueShare(request.maxAllowedRewardShare)),
+        maxAllowedRewardShare: BigInt(
+          getRevenueShare(request.maxAllowedRewardShare, RevShareType.MAX_ALLOWED_REWARD_SHARE),
```
DracoLi, line 152: Please don't mix changes that is not part of the rc1 release plans with changes that is included. Combining them does not allow us to not include these in the next release if its not ready.
This is fine for now, but in the future, this needs to be separate PR.

packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -668,7 +675,7 @@ export class IPAssetClient {
       const contractCall = () => {
         return this.licenseAttachmentWorkflowsClient.mintAndRegisterIpAndAttachPilTerms(object);
       };
-      const rsp = await this.commonRegistrationHandler({
+      const rsp = await this.handleRegistrationTransaction({
```
DracoLi, line 678: lets change it to `handleRegistrationWithFees`

packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -1969,10 +1984,16 @@ export class IPAssetClient {
       encodedTxs,
     });
     if (receipt) {
-      const { ipId, tokenId } = this.getIpIdAndTokenIdsFromEvent(receipt)[0];
-      return { txHash, ipId, tokenId, receipt };
-    } else {
-      return { txHash };
+      const event = this.getIpIdAndTokenIdsFromEvent(receipt)?.[0];
```
DracoLi, line 1987: why is this needed? do you need to fix the return type of `getIpIdAndTokenIdsFromEvent`? i don't think it indicates it can return undefined right now.

packages/core-sdk/src/types/common.ts
```diff
@@ -67,3 +67,21 @@ export type ValidatedLicensingConfig = {
  * Will be converted to bigint for contract calls.
  */
 export type TokenAmountInput = bigint | number;
+
+/**
+ * The type of revenue share.
```
DracoLi, line 72: please indicate what this is used for. Seems to be only used for error msg.

packages/core-sdk/src/types/common.ts
```diff
@@ -67,3 +67,21 @@ export type ValidatedLicensingConfig = {
...
+ */
+export enum RevShareType {
+  /**
+   * The commercial revenue share.
```
DracoLi, line 76: think these docs are not needed. it doesn't have any new info that the var does of `COMMERCIAL_REVENUE_SHARE` does not tell you already.

packages/core-sdk/test/unit/resources/ipAsset.test.ts
```diff
@@ -661,6 +661,9 @@ describe("Test IpAssetClient", () => {
       sinon.stub(ipAssetClient.licenseRegistryReadOnlyClient, "getRoyaltyPercent").resolves({
         royaltyPercent: 100,
       });
+      // Because registerDerivative doesn't call trigger parseTxIpRegisteredEvent, but the `handleRegistrationTransaction`
```
DracoLi, line 664: `registerDerivative`?

packages/core-sdk/src/resources/ipAsset.ts
bonnie57: This is because not every transition will trigger this event. If you attempt to query an event that does not exist, it will throw an error.

packages/core-sdk/test/unit/resources/ipAsset.test.ts
bonnie57: This is the reason I changed the `handleRegistrationTransaction` response. Because the method doesn't trigger `IPRegistered` event

packages/core-sdk/src/resources/group.ts
DracoLi: Please ensure I've reviewed any new items you've added to rc.1. I did not know this was added to this release. 
As for if this needs a separate PR, yes it does. This change is not related to any other change in this PR. Each PR should do one thing or a few changes of a similar issue. Small PRs are normal and very much fine.

packages/core-sdk/src/resources/ipAsset.ts
DracoLi: :+1:, i see this is for `registerDerivative`

DracoLi (approved):

packages/core-sdk/test/unit/resources/ipAsset.test.ts
DracoLi: The change for `registerDerivative` is not included in the description, so I didn't know it was included in this PR.

bonnie57: Sorry, I forgot it.

PR #430, opened 2025-02-20
More Robust Type Extraction

cryptoryda:
- Ensured `parameter.type` is checked before calling `.getText()`

- Used `path.join()` instead of hardcoded slashes for better compatibility across Windows and Unix-based systems

### DIFF
--- a/packages/react-sdk/generator/index.js
+++ b/packages/react-sdk/generator/index.js
@@ -1,165 +1,142 @@
-const fs = require("fs");
+const fs = require("fs").promises; // Use promises for better async handling
 const path = require("path");
 const { exec } = require("child_process");
-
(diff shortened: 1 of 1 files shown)


bonnie57 (commented):
Hi @Cryptoryda,  

Regarding your suggested improvements:  
- Since we’re working with `__dirname` (which is already an absolute path) and need to navigate up directory levels to reach the target, `path.resolve()` is the better choice.  
- For `parameter.type`, in this case, it will never be empty, so there’s no need to check for it.  

For now, I’ll go ahead and close this PR, but please don’t hesitate to reach out if you have any further suggestions or questions. Your contributions are always welcome!  

Best regards,  
Bonnie

PR #432, opened 2025-02-20
Origin/bp/integration/423

bpolania:
## Description
Integration tests for IP Infringement.

Added the following tests cases on `Dispute resolution`:

```
should tag a single IP as infringing without using multicall (10919ms)
should tag multiple IPs as infringing using multicall (29346ms)
should tag multiple IPs without multicall when specified (27993ms)
should fail when trying to tag with invalid dispute ID (291ms)
```

### DIFF
--- a/packages/core-sdk/test/integration/dispute.test.ts
+++ b/packages/core-sdk/test/integration/dispute.test.ts
@@ -207,7 +207,9 @@ describe("Dispute Functions", () => {
         transport: http(RPC),
         account: privateKeyToAccount(process.env.JUDGE_PRIVATE_KEY as Address),
       });
+    });
 
+    beforeEach(async function (this: Mocha.Context) {
       // Setup NFT collection
       const txData = await clientA.nftClient.createNFTCollection({
         name: "test-collection",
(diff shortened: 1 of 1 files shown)


packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -337,8 +339,230 @@ describe("Dispute Functions", () => {
...
+
+      // Verify dispute state using direct contract call
+      const disputeState = await publicClient.readContract({
+        address: disputeModuleAddress[aeneid],
+        abi: disputeModuleAbi,
+        functionName: "disputes",
+        args: [disputeId],
+      });
+
+      expect(disputeState[6]).to.equal(disputeState[5]); // currentTag should equal targetTag
```
DracoLi, line 363: what is disputeState[6] and disputeState[5], its not very clear. what response data format is `disputeState`?

bpolania: If you look at the `DisputeModule.sol` contract, the returned values are ordered:

```
function disputes(uint256 disputeId) ... returns (
    address targetIpId,          // [0]
    address disputeInitiator,    // [1]
    uint256 disputeTimestamp,    // [2]
    address arbitrationPolicy,   // [3]
    bytes32 disputeEvidenceHash, // [4]
    bytes32 targetTag,          // [5]
    bytes32 currentTag,         // [6]
    uint256 infringerDisputeId  // [7]
)
```

So, comparing ndices [5] and [6]  we are comparing `targetTag` and `currentTag`. So, when a dispute is judged successfully, the `currentTag` should equal the `targetTag`. This check verifies that the judgment was recorded correctly.

But I see your point, to improve readibility I could destructure these values, it'd be something like this:

```
const {
  targetTag,
  currentTag
} = await publicClient.readContract({
  address: disputeModuleAddress[aeneid],
  abi: disputeModuleAbi,
  functionName: 'disputes',
  args: [disputeId]
});
expect(currentTag).to.equal(targetTag);
```

I will update the tests.

DracoLi: :+1:, the explanation was very helpful. Yes, if we can either update the code so its more self-explaining or add some comments or both, that'd be great. Since we use this logic a few time, maybe a util method can also improve the readability.

DracoLi (approved):

PR #435, opened 2025-02-24
Add batched claim all revenue method

bonnie57:
## Description
- Add `batchClaimAllRevenue` method
- Add `royalty` module unit tests

### DIFF
--- a/packages/core-sdk/package.json
+++ b/packages/core-sdk/package.json
@@ -18,8 +18,8 @@
   "scripts": {
     "build": "pnpm run fix && preconstruct build",
     "test": "pnpm run test:unit",
-    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/*.test.ts'",
-    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/*.test.ts' --timeout 300000",
+    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/royalty.test.ts'",
+    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/royalty.test.ts' --timeout 300000",
     "fix": "pnpm run format:fix && pnpm run lint:fix",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
--- a/packages/core-sdk/package.json
+++ b/packages/core-sdk/package.json
@@ -18,8 +18,8 @@
   "scripts": {
     "build": "pnpm run fix && preconstruct build",
     "test": "pnpm run test:unit",
-    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/royalty.test.ts'",
-    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/royalty.test.ts' --timeout 300000",
+    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/*.test.ts'",
+    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/*.test.ts' --timeout 300000",
     "fix": "pnpm run format:fix && pnpm run lint:fix",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
--- a/packages/core-sdk/src/resources/royalty.ts
+++ b/packages/core-sdk/src/resources/royalty.ts
@@ -1,11 +1,22 @@
-import { Address, encodeFunctionData, erc20Abi, Hex, PublicClient, zeroAddress } from "viem";
+import {
+  Address,
+  encodeFunctionData,
+  erc20Abi,
+  Hex,
+  PublicClient,
+  TransactionReceipt,
+  zeroAddress,
+} from "viem";
 
 import { handleError } from "../utils/errors";
 import {
+  BatchClaimAllRevenueRequest,
+  BatchClaimAllRevenueResponse,
   ClaimableRevenueRequest,
   ClaimableRevenueResponse,
   ClaimAllRevenueRequest,
   ClaimAllRevenueResponse,
+  ClaimedToken,
   PayRoyaltyOnBehalfRequest,
   PayRoyaltyOnBehalfResponse,
   TransferClaimedTokensFromIpToWalletParams,
@@ -15,6 +26,7 @@ import {
   IpAssetRegistryClient,
   IpRoyaltyVaultImplEventClient,
   IpRoyaltyVaultImplReadOnlyClient,
+  IpRoyaltyVaultImplRevenueTokenClaimedEvent,
   Multicall3Client,
   RoyaltyModuleClient,
   RoyaltyWorkflowsClient,
@@ -74,15 +86,7 @@ export class RoyaltyClient {
       txHashes.push(txHash);
 
       // determine if the claimer is an IP owned by the wallet
-      const isClaimerIp = await this.ipAssetRegistryClient.isRegistered({
-        id: claimer,
-      });
-      const ipAccount = new IpAccountImplClient(this.rpcClient, this.wallet, claimer);
-      let ownsClaimer = claimer === this.walletAddress;
-      if (isClaimerIp) {
-        const ipOwner = await ipAccount.owner();
-        ownsClaimer = ipOwner === this.walletAddress;
-      }
+      const { ownsClaimer, isClaimerIp, ipAccount } = await this.validateClaimer(claimer);
 
       // if wallet does not own the claimer then we cannot auto claim or unwrap
       if (!ownsClaimer) {
@@ -104,7 +108,7 @@ export class RoyaltyClient {
       } else if (autoUnwrapIp && this.walletAddress === claimer) {
         // if the claimer is the wallet, then we can unwrap any claimed WIP tokens
         for (const { token, amount } of claimedTokens) {
-          if (token !== WIP_TOKEN_ADDRESS) {
+          if (token !== this.wrappedIpClient.address) {
             continue;
           }
           const hash = await this.wrappedIpClient.withdraw({
@@ -117,6 +121,108 @@ export class RoyaltyClient {
       return { receipt, claimedTokens, txHashes };
     } catch (error) {
       handleError(error, "Failed to claim all revenue");
+    }
+  }
+
+  /**
+   * Automatically batch claims all revenue from the child IPs of multiple ancestor IPs.
+   * if multicall is disabled, it will call @link{claimAllRevenue} for each ancestor IP.
+   * Then transfer all claimed tokens to the wallet if the wallet owns the IP or is the claimer.
+   * If claimed token is WIP, it will also be converted back to IP.
+   */
+  public async batchClaimAllRevenue(
+    request: BatchClaimAllRevenueRequest,
+  ): Promise<BatchClaimAllRevenueResponse> {
+    try {
+      // if the number of ancestor IPs is 1 or if multicall is disabled, then just call claimAllRevenue.
+      const useMulticallWhenPossible = request.options?.useMulticallWhenPossible !== false;
+      if (request.ancestorIps.length === 1 || !useMulticallWhenPossible) {
+        const txHashes: Hex[] = [];
+        const receipts: TransactionReceipt[] = [];
+        const claimedTokens: ClaimedToken[] = [];
+        for (const ancestorIp of request.ancestorIps) {
+          const result = await this.claimAllRevenue({
+            ...ancestorIp,
+            ancestorIpId: validateAddress(ancestorIp.ipId),
+            claimOptions: request.claimOptions,
+          });
+          txHashes.push(...result.txHashes);
+          if (result.receipt) {
+            receipts.push(result.receipt);
+          }
+          if (result.claimedTokens) {
+            claimedTokens.push(...result.claimedTokens);
+          }
+        }
+        return { txHashes, receipts, claimedTokens };
+      }
+
+      // Otherwise, we need to batch the calls into a single multicall
+      const encodedTxs = request.ancestorIps.map(
+        ({ ipId, claimer, childIpIds, royaltyPolicies, currencyTokens }) => {
+          const claim = {
+            ancestorIpId: validateAddress(ipId),
+            claimer: validateAddress(claimer),
+            childIpIds: validateAddresses(childIpIds),
+            royaltyPolicies: validateAddresses(royaltyPolicies),
+            currencyTokens: validateAddresses(currencyTokens),
+          };
+          return this.royaltyWorkflowsClient.claimAllRevenueEncode(claim).data;
+        },
+      );
+      const txHashes: Hex[] = [];
+      const txHash = await this.royaltyWorkflowsClient.multicall({ data: encodedTxs });
+      txHashes.push(txHash);
+      const receipt = await this.rpcClient.waitForTransactionReceipt({ hash: txHash });
+      const claimedTokens =
+        this.ipRoyaltyVaultImplEventClient.parseTxRevenueTokenClaimedEvent(receipt);
+      // Aggregate claimed tokens by claimer and token address
+      const aggregatedTokens = claimedTokens.reduce<
+        Record<string, IpRoyaltyVaultImplRevenueTokenClaimedEvent>
+      >((acc, curr) => {
+        const key = `${curr.claimer}_${curr.token}`;
+        if (!acc[key]) {
+          acc[key] = { ...curr };
+        } else {
+          acc[key].amount += curr.amount;
+        }
+        return acc;
+      }, {});
+      const claimers = [...new Set(request.ancestorIps.map(({ claimer }) => claimer))];
+      const autoTransfer = request.claimOptions?.autoTransferAllClaimedTokensFromIp !== false;
+      const autoUnwrapIp = request.claimOptions?.autoUnwrapIpTokens !== false;
+      for (const claimer of claimers) {
+        const { ownsClaimer, isClaimerIp, ipAccount } = await this.validateClaimer(claimer);
+        if (!ownsClaimer) {
+          continue;
+        }
+        // transfer claimed tokens from IP to wallet if wallet owns IP
+        if (autoTransfer && isClaimerIp && ownsClaimer) {
+          const hashes = await this.transferClaimedTokensFromIpToWallet({
+            ipAccount,
+            autoUnwrapIp,
+            claimedTokens: Object.values(aggregatedTokens).filter(
+              (item) => item.claimer === claimer,
+            ),
+          });
+          txHashes.push(...hashes);
+        } else if (autoUnwrapIp && this.walletAddress === claimer) {
+          // if the claimer is the wallet, then we can unwrap any claimed WIP tokens
+          for (const { token, amount } of claimedTokens) {
+            if (token !== WIP_TOKEN_ADDRESS) {
+              continue;
+            }
+            const hash = await this.wrappedIpClient.withdraw({
+              value: amount,
+            });
+            txHashes.push(hash);
+            await this.rpcClient.waitForTransactionReceipt({ hash });
+          }
+        }
+      }
+      return { receipts: [receipt], claimedTokens, txHashes };
+    } catch (error) {
+      handleError(error, "Failed to batch claim all revenue");
     }
   }
 
@@ -262,5 +368,18 @@ export class RoyaltyClient {
       await transferToken(token, amount);
     }
     return txHashes;
+  }
+
+  private async validateClaimer(claimer: Address) {
+    const isClaimerIp = await this.ipAssetRegistryClient.isRegistered({
+      id: claimer,
+    });
+    const ipAccount = new IpAccountImplClient(this.rpcClient, this.wallet, claimer);
+    let ownsClaimer = claimer === this.walletAddress;
+    if (isClaimerIp) {
+      const ipOwner = await ipAccount.owner();
+      ownsClaimer = ipOwner === this.walletAddress;
+    }
+    return { ownsClaimer, isClaimerIp, ipAccount };
   }
 }
--- a/packages/core-sdk/src/resources/royalty.ts
+++ b/packages/core-sdk/src/resources/royalty.ts
@@ -377,7 +377,7 @@ export class RoyaltyClient {
 
   private async unwrapWipTokens(claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[]) {
     const wipToken = claimedTokens.find((token) => token.token === WIP_TOKEN_ADDRESS);
-    if (!wipToken) {
+    if (!wipToken || wipToken.amount <= 0n) {
       return;
     }
     const hash = await this.wrappedIpClient.withdraw({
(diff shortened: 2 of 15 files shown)


packages/core-sdk/src/resources/royalty.ts
```diff
@@ -57,81 +57,179 @@ export class RoyaltyClient {
...
+        },
+      );
+      const txHashes: Hex[] = [];
+      const txHash = await this.royaltyWorkflowsClient.multicall({ data: encodedTxs });
+      txHashes.push(txHash);
+      const receipt = await this.rpcClient.waitForTransactionReceipt({ hash: txHash });
+      const claimedTokens =
+        this.ipRoyaltyVaultImplEventClient.parseTxRevenueTokenClaimedEvent(receipt);
+      // Aggregate claimed tokens by claimer and token address
+      const aggregatedClaimedTokens = Object.values(
```
DracoLi, line 176: I think this aggregate logic is hard to follow. can we simplify this so its easier to read and understand? 
I think we can use an array instead of reduce and then construct an object with key as the claimer address and value as the array of claimed tokens.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -57,81 +57,179 @@ export class RoyaltyClient {
...
+          txHashes.push(...hashes);
         }
       }
-      return { receipt, claimedTokens, txHashes };
+      if (ownsClaimerCount === 0) {
+        return { receipts: [receipt], txHashes };
+      }
+      return {
+        receipts: [receipt],
+        claimedTokens: aggregatedClaimedTokens,
```
DracoLi, line 225: i think for the batch call, we want to return an object here. Key is the claimer, and value is the total claimed tokens for that claimer.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -287,19 +360,38 @@ export class RoyaltyClient {
...
+    let ownsClaimer = claimer === this.walletAddress;
+    if (isClaimerIp) {
+      const ipOwner = await ipAccount.owner();
+      ownsClaimer = ipOwner === this.walletAddress;
+    }
+    return { ownsClaimer, isClaimerIp, ipAccount };
+  }
+
+  private async unwrapWIPTokens(claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[]) {
+    const txHashes: Hex[] = [];
```
DracoLi, line 384: can we make sure we never have `claimedTokens` with multiple wip records? We should never get there as it should already be aggregated. We then should update this to return return a single txHash.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -287,19 +360,38 @@ export class RoyaltyClient {
...
+    const ipAccount = new IpAccountImplClient(this.rpcClient, this.wallet, claimer);
+    let ownsClaimer = claimer === this.walletAddress;
+    if (isClaimerIp) {
+      const ipOwner = await ipAccount.owner();
+      ownsClaimer = ipOwner === this.walletAddress;
+    }
+    return { ownsClaimer, isClaimerIp, ipAccount };
+  }
+
+  private async unwrapWIPTokens(claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[]) {
```
DracoLi, line 383:
```suggestion
  private async unwrapWipTokens(claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[]) {
```
Consistent with other function names for wip.

packages/core-sdk/src/types/resources/royalty.ts
```diff
@@ -185,6 +74,46 @@ export type ClaimAllRevenueRequest = {
...
+  ancestorIps: {
+    /** The address of the ancestor IP from which the revenue is being claimed. */
+    ipId: Address;
+    /**
+     * The address of the claimer of the currency (revenue) tokens.
+     *
+     * This is normally the ipId of the ancestor IP if the IP has all royalty tokens.
+     * Otherwise, this would be the address that is holding the ancestor IP royalty tokens.
+     */
+    claimer: Address;
```
DracoLi, line 88: can we try to reuse `ClaimAllRevenueRequest` properties so we don't have to duplicate the jsdoc?
So could be something like:
```
ancestorIps: Omit<ClaimAllRevenueRequest, 'ancestorIpId' | "wipOptions"> & { 
 /** The address of the ancestor IP from which the revenue is being claimed. */
 ipId: Address
}
```

packages/core-sdk/src/types/resources/royalty.ts
```diff
@@ -185,6 +74,46 @@ export type ClaimAllRevenueRequest = {
...
+    /**
+     * The addresses of the royalty policies, where
+     * royaltyPolicies[i] governs the royalty flow for childIpIds[i].
+     */
+    royaltyPolicies: Address[];
+    /** The addresses of the currency tokens in which royalties will be claimed */
+    currencyTokens: Address[];
+  }[];
+  /** The deadline for the transaction. */
+  deadline?: bigint | number | string;
```
DracoLi, line 100: add a note on the default for this if not provided

packages/core-sdk/src/types/resources/royalty.ts
```diff
@@ -199,6 +128,6 @@ export type ClaimAllRevenueResponse = {
 
 export type TransferClaimedTokensFromIpToWalletParams = {
   ipAccount: IpAccountImplClient;
-  skipUnwrapIp: boolean;
+  autoUnwrapIp: boolean;
```
DracoLi, line 131: i don't think this is needed anymore

packages/core-sdk/test/integration/royalty.test.ts
```diff
@@ -258,7 +263,7 @@ describe("Royalty Functions", () => {
               derivativesApproval: false,
               derivativesReciprocal: true,
               derivativeRevCeiling: 0n,
-              currency: WIP_TOKEN_ADDRESS,
+              currency: wrappedIpAddress[aeneid],
```
DracoLi, line 266: lets keep using `WIP_TOKEN_ADDRESS`, it will never change and its easier to use.

packages/core-sdk/test/integration/royalty.test.ts
```diff
@@ -330,4 +335,215 @@ describe("Royalty Functions", () => {
       expect(ret.claimedTokens![0].amount).to.equal(120n);
     });
   });
+
+  describe("BatchClaimAllRevenue With WIP", () => {
```
DracoLi, line 339: lets move this inside of the existing `ClaimAllRevenue With WIP` test so we don't have to redo some of the setup work. We can have two inner describes in there, one for batching and one without batching.
Goal here is to reuse setup work for integration tests.

packages/core-sdk/package.json
```diff
@@ -18,8 +18,8 @@
   "scripts": {
     "build": "pnpm run fix && preconstruct build",
     "test": "pnpm run test:unit",
-    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/*.test.ts'",
-    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/*.test.ts' --timeout 300000",
+    "test:unit": "TS_NODE_PROJECT='./tsconfig.test.json' c8 --all --src ./src mocha -r ts-node/register './test/unit/**/royalty.test.ts'",
+    "test:integration": "TS_NODE_PROJECT='./tsconfig.test.json' mocha -r ts-node/register './test/integration/**/royalty.test.ts' --timeout 300000",
```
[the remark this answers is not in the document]
bonnie57: Sorry, Forgot to commit.

packages/core-sdk/src/resources/royalty.ts
bonnie57: I observed that "wip" is consistently used throughout the code, with the exception of the WIPTokenClient class. However, the inline annotations use "WIP" instead. Do we need to ensure consistency in this regard?

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: I noticed that `WIP_TOKEN_ADDRESS` is being used to store the wrapper IP contract address. However, I recommend using a **single source of truth** in `generated.ts` instead. This approach ensures that if the contract address changes, we won't need to manually update `WIP_TOKEN_ADDRESS`. Moreover, we can directly fetch the contract address from `generated.ts`, eliminating the need for an additional variable.

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: I have tried, but it's difficult to reuse data because data are used by `ClaimAllRevenue `.

packages/core-sdk/src/resources/royalty.ts
bonnie57: To aggregate the amount, using `reduce` is a suitable method. I'm unclear about why you find it difficult to follow—are you referring to the logic being complex or hard to understand?

As for the key, here's a solution: when claims have the same claimer but involve different tokens, we cannot aggregate the amounts. Thus, the key should be a combination of `claimer` + `token`, which guarantees uniqueness.

packages/core-sdk/src/resources/royalty.ts
bonnie57: This is a nice-to-have feature, and I actually thought about it as well. However, as I mentioned, it can be challenging to distinguish claims solely by the claimer's address.

Regarding `batchClaimAllRevenue` and `claimAllRevenue`, we should strive to maintain a consistent structure between the two. This will improve code readability and ensure a unified approach across the codebase.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -57,81 +57,179 @@ export class RoyaltyClient {
...
+  ): Promise<BatchClaimAllRevenueResponse> {
+    try {
+      // if the number of ancestor IPs is 1 or if multicall is disabled, then just call claimAllRevenue.
+      const useMulticallWhenPossible = request.options?.useMulticallWhenPossible !== false;
+      if (request.ancestorIps.length === 1 || !useMulticallWhenPossible) {
+        const txHashes: Hex[] = [];
+        const receipts: TransactionReceipt[] = [];
+        const claimedTokens: ClaimedToken[] = [];
+        for (const ancestorIp of request.ancestorIps) {
+          const result = await this.claimAllRevenue({
```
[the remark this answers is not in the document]
bonnie57: Changed it, please review it again.

packages/core-sdk/test/integration/royalty.test.ts
DracoLi: okay, if difficult to reuse, then this looks good.

packages/core-sdk/test/integration/royalty.test.ts
DracoLi: yep that case, you can set the value of WIP_TOKEN_ADDRESS to the value in `generated.ts`. I don't like needing to import both wrappedIpAddress and the chain id when the address is the same.

packages/core-sdk/src/resources/royalty.ts
DracoLi: yes you can. key is claimer, value is an array of tokens and their amounts.

packages/core-sdk/src/resources/royalty.ts
DracoLi: i disagree, i think the goal is to provide the response data in a format that make sense for the endpoint and for the users. I think aggregated claim amount per claimer makes the most sense for this endpoint so lets follow that for now.

packages/core-sdk/test/integration/royalty.test.ts
DracoLi: i actually recommend using `WIP_TOKEN_ADDRESS` everywhere and not use the generated values.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -135,6 +120,122 @@ export class RoyaltyClient {
...
+      const txHashes: Hex[] = [];
+      const receipts: TransactionReceipt[] = [];
+      const claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[] = [];
+      // if the number of ancestor IPs is 1 or if multicall is disabled, then just call claimAllRevenue.
+      const useMulticallWhenPossible = request.options?.useMulticallWhenPossible !== false;
+      if (request.ancestorIps.length === 1 || !useMulticallWhenPossible) {
+        for (const ancestorIp of request.ancestorIps) {
+          const result = await this.claimAllRevenue({
+            ...ancestorIp,
+            ancestorIpId: validateAddress(ancestorIp.ipId),
```
DracoLi, line 142: not really needed to validate here since `claimAllRevenue` validated again.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -135,6 +120,122 @@ export class RoyaltyClient {
...
+            }
+            return acc;
+          },
+          {},
+        ),
+      );
+      const claimers = [...new Set(request.ancestorIps.map(({ claimer }) => claimer))];
+      const autoTransfer = request.claimOptions?.autoTransferAllClaimedTokensFromIp !== false;
+      const autoUnwrapIp = request.claimOptions?.autoUnwrapIpTokens !== false;
+      let ownsClaimerCount = 0;
```
DracoLi, line 195: why is this needed here? in the non-batched call, i think we return the `claimedTokens` even if wallet does not own claimer.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -287,19 +363,35 @@ export class RoyaltyClient {
...
+    let ownsClaimer = claimer === this.walletAddress;
+    if (isClaimerIp) {
+      const ipOwner = await ipAccount.owner();
+      ownsClaimer = ipOwner === this.walletAddress;
+    }
+    return { ownsClaimer, isClaimerIp, ipAccount };
+  }
+
+  private async unwrapWipTokens(claimedTokens: IpRoyaltyVaultImplRevenueTokenClaimedEvent[]) {
+    const wipToken = claimedTokens.find((token) => token.token === WIP_TOKEN_ADDRESS);
```
DracoLi, line 387: lets be more defensive here. We should throw an error if there are multiple wip tokens or we can aggregate the values. I think aggregate might be a bit better?

packages/core-sdk/src/resources/royalty.ts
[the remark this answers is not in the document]
DracoLi: looks good now, thx

packages/core-sdk/src/resources/royalty.ts
DracoLi: not a blocker for this PR. i can also take a look as some potential refactor options later.

packages/core-sdk/src/resources/royalty.ts
DracoLi: yes, WIP should be in docs. for variables, I think we can keep it as `wip` if starting name, or`Wip`.

packages/core-sdk/src/resources/royalty.ts
bonnie57: ClaimedToken does not include entries that share the same token address and claimer. This is because ClaimedToken is an aggregated collection of ClaimedTokens, grouped by the same token address and claimer.

It meets your requirements, as it is an array with the type `IpRoyaltyVaultImplRevenueTokenClaimedEvent`. Please refer to the code for further details.

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: I disagree with hardcoding `WIP_TOKEN_ADDRESS` because:

1. **Single Source of Truth**: Using `generated.ts` ensures consistency and avoids manual updates if the address changes.  
2. **Consistency**: Other addresses are dynamically fetched from `generated.ts`—`WIP_TOKEN_ADDRESS` should follow the same pattern.  
3. **Maintainability**: Hardcoding increases maintenance overhead, especially for multi-chain or multi-environment support.  
4. **Flexibility**: Fetching from `generated.ts` adapts better to future changes.  

Let’s keep all addresses consistent by sourcing them from `generated.ts`.

packages/core-sdk/src/resources/royalty.ts
bonnie57: While there is only one WIP token, the system might interact with other tokens, such as `ERC20`. Currently, the `unwrapWipTokens` method only focuses on handling `WIP`.  I don't want to interrupt user interaction when the token is another token instead of WIP.

packages/core-sdk/test/integration/royalty.test.ts
DracoLi: I've considered your points, and I'm making a call to migrate to use `WIP_TOKEN_ADDRESS`. You can update the value of `WIP_TOKEN_ADDRESS` so it takes it from the generated values instead, I'm also fine with just hard coding the values. This PR will not be approved unless those are changed.

packages/core-sdk/src/resources/royalty.ts
DracoLi: I've made it clear you can aggregate the wip sum and unwrap. please make the change.

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: Thank you for your feedback and for making a clear decision on using `WIP_TOKEN_ADDRESS`. While I still believe sourcing values from `generated.ts` offers better consistency and maintainability, I respect your call and will update the code to align with the agreed approach.

Regarding the PR process, I see it as a collaborative way for us to discuss and refine changes, especially since it’s just the two of us working on this. For example, You’d agree with renaming `skipTransfer` and `skipUnwrapIp` to `autoTransfer` and `autoUnwrapIp`, so I made those changes. If you’d prefer to discuss such changes more explicitly beforehand, I’m happy to adjust my workflow.

packages/core-sdk/src/resources/royalty.ts
bonnie57: I think there might be some missing context here. The `claimedTokens` data is already aggregated, and its structure looks like this:

```json
[
  { token: WIP_TOKEN_ADDRESS, amount: 2n, claimer: walletAddress },
  { token: ERC20, amount: 0n, claimer: walletAddress }
]
```
This means `claimedTokens` does not represent `multiple WIP tokens`. Additionally, if the token is ERC20, should we throw an error? Let me know your thoughts!

packages/core-sdk/test/integration/royalty.test.ts
DracoLi: I think if the change is related to the PR then including your suggested changes it in the PR along with any reasons for the change in the PR description is a great way to get feedback. 
But I don't view changing `WIP_TOKEN_ADDRESS` here to `wrappedIpAddress[aeneid]` is required for the PR to add `batchClaimAllRevenue`. For those changes, its good to either start a discussion elsewhere, or putting a PR with just that refactor and get feedback.

packages/core-sdk/src/resources/royalty.ts
DracoLi: since `unwrapWipTokens` is a util, its not guaranteed that that's passed in is aggregated. It could be right now but someone can use it elsewhere and pass an array with two WIP records.
You can do the following:
- Make sure there's no two WIP token records in `claimedTokens`, otherwise throw error. I've mentioned nothing about throwing an error if there's another erc20 in there.
- Aggregate WIP inside this function, to ensure we can never get two wip records.

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -57,76 +56,62 @@ export class RoyaltyClient {
...
     this.ipAccountClient = new IPAccountClient(rpcClient, wallet);
     this.multicall3Client = new Multicall3Client(rpcClient, wallet);
     this.wrappedIpClient = new WrappedIpClient(rpcClient, wallet);
     this.rpcClient = rpcClient;
     this.wallet = wallet;
     this.walletAddress = wallet.account!.address;
   }
-
+  /**
+   * Claims all revenue from the child IPs of an ancestor IP, then transfer.
```
DracoLi, line 68: :+1: thx for adding this

packages/core-sdk/src/types/resources/royalty.ts
```diff
@@ -185,20 +78,41 @@ export type ClaimAllRevenueRequest = {
     autoUnwrapIpTokens?: boolean;
   };
 };
+export type BatchClaimAllRevenueRequest = WithClaimOptions & {
+  /** The ancestor IPs from which the revenue is being claimed. */
+  ancestorIps: (Omit<ClaimAllRevenueRequest, "ancestorIpId" | "wipOptions"> & {
```
DracoLi, line 83: should this be `"ancestorIpId" | "claimOptions"`? There is no `wipOptions` in ClaimAllRevenueRequest

packages/core-sdk/src/types/resources/royalty.ts
```diff
@@ -185,20 +78,41 @@ export type ClaimAllRevenueRequest = {
...
 
 export type ClaimAllRevenueResponse = {
   txHashes: Hash[];
-  receipt?: TransactionReceipt;
-  claimedTokens?: ClaimedToken[];
+  receipt: TransactionReceipt;
+  /**
+   * Aggregate list of all tokens claimed across all transactions in the batch.
+   * Events are aggregated by unique combinations of claimer and token addresses,
+   * summing up the amounts for the same claimer-token pairs.
```
DracoLi, line 110: :+1:, very useful.

packages/core-sdk/test/integration/royalty.test.ts
DracoLi, line 266: this change still need to be reverted in this PR.

packages/core-sdk/test/integration/royalty.test.ts
```diff
@@ -330,4 +335,215 @@ describe("Royalty Functions", () => {
...
+      });
+      ipB3 = ret5?.[0].ipId!;
+      const balance = await client.wipClient.balanceOf(TEST_WALLET_ADDRESS);
+      await client.wipClient.withdraw({
+        amount: balance,
+        txOptions: { waitForTransaction: true },
+      });
+    });
+
+    it("should batch claim all revenue and convert WIP back to IP", async () => {
```
DracoLi, line 525: `and convert WIP back to IP` this part isn't really tested in your assertions here. If possible, we should validate this, otherwise just remove this part so its clear we are not validating it.

packages/core-sdk/test/integration/royalty.test.ts
```diff
@@ -330,4 +335,215 @@ describe("Royalty Functions", () => {
...
+      const result = await client.royalty.batchClaimAllRevenue({
+        ancestorIps: [
+          {
+            ipId: ipA,
+            claimer: ipA,
+            childIpIds: [ipA1, ipA2],
+            royaltyPolicies: [royaltyPolicyLapAddress[aeneid], royaltyPolicyLapAddress[aeneid]],
```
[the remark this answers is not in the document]
bonnie57: How about `WIP` address?  Do we need to export it?

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: Agree. I noticed `ClaimAllRevenue` also doesn't validate, I will also remove it.

packages/core-sdk/test/integration/royalty.test.ts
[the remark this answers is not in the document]
DracoLi: we already export `WIP_CONTRACT_ADDRESS` which is same for all supported chains. We don't need to export `wrappedIpAddress`.

packages/core-sdk/test/integration/royalty.test.ts
bonnie57: I agree with you. It's dynamic gas fees. I have removed the description of both.

DracoLi (approved):
LGTM

PR #437, opened 2025-02-25
Update ip metadata url

bonnie57:


packages/core-sdk/src/resources/ipAccount.ts
```diff
@@ -150,4 +158,41 @@ export class IPAccountClient {
...
+        txOptions: {
+          ...txOptions,
+          encodedTxDataOnly: false,
+        },
+      });
+      return txHash!;
+    } catch (error) {
+      handleError(
+        new Error(
+          (error as Error).message
```
DracoLi, line 190: we should try not recreate the original error unless there's no good alternative. The original error contains more information than just the message. Also the error might not be type `Error` so that needs to be handled as well. Furthermore, this does not retain the original stacktrace, which needs to be fixed.

I'd rather just retain the original error, instead of doing some manual error msg clean up for the `Failed to execute the IP Account transaction:` text.

PR #439, opened 2025-02-26
feat: Add dispute assertion to dispute client

DracoLi:
## Description
Counters a dispute that was raised by another party on an IP using counter evidence that is uploaded to IPFS.

This method can only be called by the IP's owner to counter a dispute by providing counter evidence. The counter evidence (e.g., documents, images) should be uploaded to IPFS, and its corresponding CID is converted internally to a hash for the request.

### Example

```jsx
const ret = await client.dispute.disputeAssertion({
  ipId, // the ipId of the disputed IP
  assertionId, // the assertionId of the dispute
  counterEvidenceCID, // CID of the counter evidence
});
```

### DIFF
--- a/packages/core-sdk/src/resources/dispute.ts
+++ b/packages/core-sdk/src/resources/dispute.ts
@@ -235,6 +235,15 @@ export class DisputeClient {
     }
   }
 
+  /**
+   * Counters a dispute that was raised by another party on an IP using counter evidence.
+   *
+   * This method can only be called by the IP's owner to counter a dispute by providing
+   * counter evidence. The counter evidence (e.g., documents, images) should be
+   * uploaded to IPFS, and its corresponding CID is converted to a hash for the request.
+   *
+   * If you only have a `disputeId`, call {@link disputeIdToAssertionId} to get the `assertionId` needed here.
+   */
   public async disputeAssertion(request: DisputeAssertionRequest): Promise<TransactionResponse> {
     try {
       const ipAccount = new IpAccountImplClient(
--- a/packages/core-sdk/src/resources/dispute.ts
+++ b/packages/core-sdk/src/resources/dispute.ts
@@ -4,6 +4,7 @@ import { handleError } from "../utils/errors";
 import {
   CancelDisputeRequest,
   CancelDisputeResponse,
+  DisputeAssertionRequest,
   RaiseDisputeRequest,
   RaiseDisputeResponse,
   ResolveDisputeRequest,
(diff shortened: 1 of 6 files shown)


packages/core-sdk/src/resources/dispute.ts
```diff
@@ -230,4 +234,49 @@ export class DisputeClient {
...
+        txHash,
+        txOptions: request.txOptions,
+        rpcClient: this.rpcClient,
+      });
+    } catch (e) {
+      handleError(e, "Failed to dispute assertion");
+    }
+  }
+
+  public async disputeIdToAssertionId(disputeId: number | bigint): Promise<Hex> {
```
bonnie57, line 276: It's better to start with `verb` about this method.

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -73,3 +73,23 @@ export type TagIfRelatedIpInfringedRequest = {
...
+
+export type DisputeAssertionRequest = {
+  /**
+   * The IP ID that is the target of the dispute.
+   */
+  ipId: Address;
+  /**
+   * The identifier of the assertion that was disputed.
+   *
+   * You can get this from the `disputeId` by calling `dispute.disputeIdToAssertionId`.
```
bonnie57, line 85: Could you please use the `@link` tag to reference `dispute.disputeIdToAssertionId`, similar to the example provided in the code below?
```
/**
 * See {@link MyClass} and [MyClass's foo property]{@link MyClass#foo}.
 */
function myFunction() {}
```

packages/core-sdk/test/unit/resources/dispute.test.ts
```diff
@@ -345,4 +326,32 @@ describe("Test DisputeClient", () => {
...
+        .stub(IpAccountImplClient.prototype, "execute")
+        .resolves(txHash);
+      const result = await disputeClient.disputeAssertion({
+        ipId,
+        assertionId: generateRandomHash(),
+        counterEvidenceCID: "QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR",
+      });
+      expect(result.txHash).equal(txHash);
+      expect(accountExecuteMock.calledOnce).to.be.true;
+    });
```
bonnie57, line 342: It seems the test error case was overlooked. When you run the unit tests, the coverage report will highlight which lines of code are not covered. Could you please add the missing test case?

bonnie57 (approved):
LGTM, except for a few minor issues.

packages/core-sdk/src/types/resources/dispute.ts
DracoLi: i'm not able to get `@link` to work across files.

PR #456, opened 2025-03-07
Add auto wip in raiseDipute and disputeAssertion

bonnie57:


packages/core-sdk/src/resources/dispute.ts
```diff
@@ -251,23 +278,194 @@ export class DisputeClient {
...
       const encodedData = this.arbitrationPolicyUmaClient.disputeAssertionEncode({
         assertionId: request.assertionId,
         counterEvidenceHash,
       });
-      const txHash = await ipAccount.execute({
-        to: encodedData.to,
-        value: 0n,
-        data: encodedData.data,
-        operation: 0,
+      // Approve ipAccount to transfer wrappedIp, cannot use ipAccount.executeBatch because msg.sender is ipAccount with same address as spender.
```
DracoLi, line 403: > cannot use ipAccount.executeBatch because msg.sender is ipAccount with same address as spender.

i think this part needs a bit more clarity. I understand we have to do this separate because we need the client wallet to approve the ipAccount to use the WIP from the client wallet to pay for the bond right? We should include that comment here.

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -251,23 +278,194 @@ export class DisputeClient {
...
+      await this.wrappedIpClient.approve({
+        spender: ipAccount.address,
+        amount: maxUint256,
```
DracoLi, line 406: we can skip this approval if ipAccount allowance is enough.

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -251,23 +278,194 @@ export class DisputeClient {
...
       });
+      const contractCall = () =>
```
DracoLi, line 408: all this new logic is only applicable if bond > 0. if bond === 0, lets skip all this and just call `disputeAssertionEncode` like before.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -110,7 +87,7 @@ describe("Dispute Functions", () => {
         cid: await generateCID(),
         targetTag: "IMPROPER_REGISTRATION",
         liveness: 2592000,
-        bond: 0,
+        bond: 1000,
```
DracoLi, line 90: lets keep this small if we can to minimize ip usage

DracoLi (approved):

DracoLi:
approved by mistake, need a few small changes.

PR #457, opened 2025-03-07
Add transferErc20 method from ipAccount

bonnie57:


packages/core-sdk/src/types/resources/ipAccount.ts
```diff
@@ -48,3 +48,16 @@ export type SetIpMetadataRequest = {
...
   txOptions?: Omit<TxOptions, "encodedTxDataOnly">;
 };
+
+export type TransferErc20Request = {
+  ipId: Address;
+  tokens: {
+    /** The address of the ERC20 token including WIP and standard ERC20. */
+    address: Address;
+    /** The amount of the ERC20 token to transfer. */
+    amount: bigint | string | number;
```
DracoLi, line 58: lets use `TokenAmountInput` for all token amount in the future. It allow us to have a consistent input type across all functions and I don't think we need string support.

packages/core-sdk/test/integration/ipAccount.test.ts
```diff
@@ -159,4 +168,78 @@ describe("IPAccount Functions", () => {
     });
     expect(txHash).to.be.a("string").and.not.empty;
   });
+
+  it.only("should successfully transfer ERC20 tokens", async () => {
```
DracoLi, line 172: remove only

packages/core-sdk/test/integration/ipAccount.test.ts
```diff
@@ -159,4 +168,78 @@ describe("IPAccount Functions", () => {
...
+    const txHash = await erc20.mint({
+      to: ipId,
+      amount: parseEther("0.002"),
+    });
+    await publicClient.waitForTransactionReceipt({
+      hash: txHash,
+    });
+    // 3. transfer wip to the ip account
+    await client.wipClient.deposit({
+      amount: parseEther("0.001"),
```
DracoLi, line 193: can we use a smaller amount here (to save IP)?. I think any amount (ie 0-10) is okay to validate right?

packages/core-sdk/test/unit/resources/ipAccount.test.ts
```diff
@@ -217,4 +221,45 @@ describe("Test IPAccountClient", () => {
       expect(result).to.equal(txHash);
     });
   });
+
+  describe("Test transferErc20", () => {
+    it("should throw error when call transferErc20 failed", async () => {
+      sinon
+        .stub(IpAccountImplClient.prototype, "executeBatch")
+        .rejects(new Error("Failed to transfer ERC20 tokens"));
+      try {
```
DracoLi, line 230: lets try to avoid using try catch if possible. I think we can convert this to.
```
await expect(...).to.throw("..")
```

packages/core-sdk/test/unit/resources/ipAccount.test.ts
```diff
@@ -217,4 +221,45 @@ describe("Test IPAccountClient", () => {
       expect(result).to.equal(txHash);
     });
   });
+
+  describe("Test transferErc20", () => {
+    it("should throw error when call transferErc20 failed", async () => {
```
DracoLi, line 226: maybe one more test case, when one of the target token address is invalid, `executeBatch` should never be called and method should throw.

DracoLi (approved):

PR #462, opened 2025-03-10
Origin/bp/disputable

bpolania:
## Description

This PR refactors the Dispute Module to implement the new dispute flow in the protocol that settles the assertions through UMA's Optimistic Oracle V3 contract. 

* Updates the test cases
* Adds the `IOOV3` and the `ArbitratioPolicyUMA` ABIs to the generated file.
* Removes the `JUDGE_PRIVATE_KEY` from the .env file


packages/core-sdk/src/abi/generated.ts
```diff
@@ -30068,3 +30068,717 @@ export class WrappedIpClient extends WrappedIpReadOnlyClient {
     };
   }
 }
+
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
bonnie57, line 30072: The file are generated. If you put code into this file, it will be overriden.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -54,6 +50,32 @@ const generateCID = async () => {
...
+  const arbitrationPolicyRawResult = await publicClient.call({
+    to: "0xfFD98c3877B8789124f02C7E8239A4b0Ef11E936" as Address,
+    data: arbitrationPolicyCallData,
+  });
+
+  const assertionId = arbitrationPolicyRawResult.data;
+
+  // First simulate the contract call to check if it would succeed
+  const { request } = await publicClient.simulateContract({
+    address: "0xABac6a158431edED06EE6cba37eDE8779F599eE4" as Address,
```
bonnie57, line 69: you can get oov3 address by calling getOov3Contract in oov3 file. It's dynamic address instead of static address.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -73,7 +95,7 @@ describe("Dispute Functions", () => {
     });
     const txHash = await clientAWalletClient.sendTransaction({
       to: walletB.address,
-      value: parseEther("0.25"),
+      value: parseEther("0.50"),
```
bonnie57, line 98: I think it is a temporary solution. We still face the gas issue when gas prices are high.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -54,6 +50,32 @@ const generateCID = async () => {
   return cidv1.toV0().toString();
 };
 
+const settleAssertion = async (disputeId: bigint): Promise<`0x${string}`> => {
```
bonnie57, line 53: It should be `Hex` instead of `0x${string}`.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -54,6 +50,32 @@ const generateCID = async () => {
...
+  const arbitrationPolicyCallData = encodeFunctionData({
+    abi: arbitrationPolicyUmaAbi,
+    functionName: "disputeIdToAssertionId",
+    args: [disputeId],
+  });
+
+  const arbitrationPolicyRawResult = await publicClient.call({
+    to: "0xfFD98c3877B8789124f02C7E8239A4b0Ef11E936" as Address,
```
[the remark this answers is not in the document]
bpolania: ArbitrationPolicyUMA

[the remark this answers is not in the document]
bonnie57: It looks like you are calling `disputeIdToAssertionId`. In the dispute module, there is a `disputeIdToAssertionId` method that you can directly call. However, your contract address is not the same as the SDK implementation. Sdk uses `0xfFD98c3877B8789124f02C7E8239A4b0Ef11E936`

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -330,39 +339,35 @@ describe("Dispute Functions", () => {
...
-        liveness: 2592000,
+        liveness: 1,
         bond: 0,
         txOptions: {
           waitForTransaction: true,
         },
       });
       disputeId = response.disputeId!;
+
+      await new Promise((resolve) => setTimeout(resolve, 3000));
```
bonnie57, line 350: Could you tell me why you need to add this?

packages/core-sdk/src/abi/generated.ts
bpolania: Fixed, using OOV3 module files

packages/core-sdk/test/integration/dispute.test.ts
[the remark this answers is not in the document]
bpolania: Fixed, using 00V3 module

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Agree. I think we need to move to a fixed account for this. I will raise this point in the conversation and submit another PR with the agreed solution,

packages/core-sdk/test/integration/dispute.test.ts
bpolania: In these cases is necessary to be sure that the disputed is expired

packages/core-sdk/test/integration/dispute.test.ts
bpolania: this is returning the hash of the transaction, these are coming as 0x strings

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -87,7 +113,7 @@ describe("Dispute Functions", () => {
         cid: await generateCID(),
         targetTag: "IMPROPER_REGISTRATION",
         liveness: 2592000,
-        bond: 1,
+        bond: 0,
```
bonnie57, line 116: Why do you need to change the number? We need a test bond for some value situations.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: waitForTransaction is an option that waits for getting a receipt. I think it can make sure the previous feature is completed. We don't need it.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -305,39 +320,35 @@ describe("Dispute Functions", () => {
...
-      const txHash = await judgeWalletClient.writeContract(request);
-      await publicClient.waitForTransactionReceipt({ hash: txHash });
+      const txHash = await settleAssertion(clientA, disputeId);
+
+      // Assert that txHash is a valid transaction hash
+      expect(txHash).to.be.a("string");
+      expect(txHash).to.match(/^0x[a-fA-F0-9]{64}$/);
+
+      // Assert the receipt comes with a success
+      const receipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
```
bonnie57, line 342: Could you add the logic into `settleAssertion`? It belongs to `settleAssertion` method.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -305,39 +320,35 @@ describe("Dispute Functions", () => {
...
-        args: [disputeId, true, "0x"],
-        account: judgeWalletClient.account!,
-      });
-      const txHash = await judgeWalletClient.writeContract(request);
-      await publicClient.waitForTransactionReceipt({ hash: txHash });
+      const txHash = await settleAssertion(clientA, disputeId);
+
+      // Assert that txHash is a valid transaction hash
+      expect(txHash).to.be.a("string");
+      expect(txHash).to.match(/^0x[a-fA-F0-9]{64}$/);
```
bonnie57, line 339: Could you let me know why you verified the txHash pattern? Does txHash maybe have other string value?

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -35,6 +41,22 @@ const generateCID = async () => {
   return cidv1.toV0().toString();
 };
 
+const settleAssertion = async (client: StoryClient, disputeId: bigint): Promise<`0x${string}`> => {
```
bonnie57, line 44: Please use `Hex` instead of `0x${string}`.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -189,7 +218,7 @@ describe("Dispute Functions", () => {
    *
    * On mainnet, disputes are judged by UMA's optimistic oracle. For testing purposes,
    * we simulate this process by setting up a whitelisted judge account that can
-   * directly set dispute judgement. The process creates a wallet client with the
+   * directly set dispute judgements. The process creates a wallet client with the
```
bonnie57, line 221: Please use `judgments` instead of `judgements,` which is the wrong word as detected by vscode.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: This is an error I will set back to 1.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: I added there to verify that the response I was getting was in effect a hash, but I think it can be removed now.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Added.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Oh, I guess I was writing on British english =), fixed!

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Done.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -54,10 +76,14 @@ describe("Dispute Functions", () => {
...
       to: walletB.address,
-      value: parseEther("0.25"),
+      value: parseEther("0.50"),
     });
     await publicClient.waitForTransactionReceipt({ hash: txHash });
 
+    // clientA approves the arbitration policyUma module to spend the some tokens
+    const mockERC20 = new WipTokenClient(publicClient, walletClient);
+    await mockERC20.approve(arbitrationPolicyUmaAddress[aeneid], maxUint256);
+
```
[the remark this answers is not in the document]
bpolania: Dead code. Removed.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: Sorry, I didn't see the ad. I mean, we should put receipt logic into `settleAssertion` method.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -347,23 +351,17 @@ describe("Dispute Functions", () => {
...
-        account: judgeWalletClient.account!,
-      });
-      const judgmentTxHash = await judgeWalletClient.writeContract(request);
-      await publicClient.waitForTransactionReceipt({ hash: judgmentTxHash });
+      // Step 1: Set judgment on an existing dispute to mark it as valid
+      const txHash = await settleAssertion(clientA, disputeId);
+
+      // Assert the receipt comes with a success
+      const receipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+      expect(receipt.status).to.equal("success");
```
bonnie57, line 364: This is the same as the previous, we shouldn't put in every case. I just put into `settleAssertion` method

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -35,6 +39,25 @@ const generateCID = async () => {
...
+    abi: ASSERTION_ABI,
+    functionName: "settleAssertion",
+    args: [assertionId],
+    account: walletClient.account,
+  });
+
+  const txHash = await walletClient.writeContract(request);
+  expect(txHash).to.be.a("string");
+
+  return txHash as Hex;
```
bonnie57, line 58: Here, you don't need to as `Hex`.   Please remove it.

packages/core-sdk/test/integration/dispute.test.ts
bonnie57: Please remove the related code, we don't need to do it manually wait.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Done.

packages/core-sdk/test/integration/dispute.test.ts
bpolania: Done.

bonnie57 (approved):
LGTM, except for two minor things:
- I don't think we need to wait manually, because waiting for the receipt is enough to wait for the transaction to complete.
```
await new Promise((resolve) => setTimeout(resolve, 3000));
```
- Awaiting receipt has already been put into `settleAssertion`, we don't need to wait receipt again in other cases when calling `settleAssertion` method.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -35,6 +39,28 @@ const generateCID = async () => {
...
 
+const settleAssertion = async (client: StoryClient, disputeId: bigint): Promise<Hex> => {
+  const arbitrationPolicyUmaClient = new ArbitrationPolicyUmaClient(publicClient, walletClient);
+  const oov3Address = await arbitrationPolicyUmaClient.oov3();
+  const assertionId = await client.dispute.disputeIdToAssertionId(disputeId!);
+
+  const { request } = await publicClient.simulateContract({
+    address: oov3Address,
+    abi: ASSERTION_ABI,
+    functionName: "settleAssertion",
```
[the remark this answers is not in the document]
DracoLi: Also some details on how the flow works. I think you mentioned `settleAssertion on OOV3 -> assertionResolvedCallback on ArbitrationPolicyUMA.sol -> setDisputeJudgement in DisputeModule.sol`. I think it'd be good to include that here for our future reference.

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -347,23 +354,17 @@ describe("Dispute Functions", () => {
...
-        args: [disputeId, true, "0x"],
-        account: judgeWalletClient.account!,
-      });
-      const judgmentTxHash = await judgeWalletClient.writeContract(request);
-      await publicClient.waitForTransactionReceipt({ hash: judgmentTxHash });
+      // Step 1: Set judgment on an existing dispute to mark it as valid
+      const txHash = await settleAssertion(clientA, disputeId);
+
+      // Assert the receipt comes with a success
+      const receipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
```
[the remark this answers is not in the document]
bpolania: Done

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -568,15 +562,11 @@ describe("Dispute Functions", () => {
...
-        functionName: "setDisputeJudgement",
-        args: [disputeId, true, "0x"],
-        account: judgeWalletClient.account!,
-      });
-      const judgmentTxHash = await judgeWalletClient.writeContract(request);
-      await publicClient.waitForTransactionReceipt({ hash: judgmentTxHash });
+      const txHash = await settleAssertion(clientA, disputeId);
+
+      // Assert the receipt comes with a success
+      const receipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
```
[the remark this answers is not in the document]
bpolania: Done

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -35,6 +39,28 @@ const generateCID = async () => {
   return cidv1.toV0().toString();
 };
 
+const settleAssertion = async (client: StoryClient, disputeId: bigint): Promise<Hex> => {
+  const arbitrationPolicyUmaClient = new ArbitrationPolicyUmaClient(publicClient, walletClient);
+  const oov3Address = await arbitrationPolicyUmaClient.oov3();
+  const assertionId = await client.dispute.disputeIdToAssertionId(disputeId!);
```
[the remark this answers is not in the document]
bpolania: Removed

packages/core-sdk/test/integration/dispute.test.ts
```diff
@@ -305,39 +320,31 @@ describe("Dispute Functions", () => {
...
+
+      // Tag derivative IP as infringing
       const results = await clientA.dispute.tagIfRelatedIpInfringed({
         infringementTags: [
           {
             ipId: childIpId,
             disputeId: disputeId,
           },
-          {
-            ipId: childIpId2,
```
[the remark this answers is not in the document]
bpolania: It was removed during the process of creating the tests, it is now restored

packages/core-sdk/test/integration/dispute.test.ts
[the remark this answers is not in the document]
bpolania: Added

bpolania:
> LGTM, except for two minor things:
> 
> * I don't think we need to wait manually, because waiting for the receipt is enough to wait for the transaction to complete.
> 
> ```
> await new Promise((resolve) => setTimeout(resolve, 3000));
> ```
> 
> * Awaiting receipt has already been put into `settleAssertion`, we don't need to wait receipt again in other cases when calling `settleAssertion` method.

Dispute assertions cannot be settle unless the liveness period has passed, so for aeneid testing purposes, very short liveness values are allowed. The test cases set a 1 second value to a dispute, and then tries to settle. This randomly fails with the Assertion not expired error.  This could be happening because chain speed or diff block. All transactions in same block have same timestamp so it's likely is reverting when both transactions fit same block. 

I discussed this matter with @Tiago and one of the proposed solutions was to allow a value of 0 for liveness, but this is not possible, so in order to guarantee that the assertion always expires and the test passes is to add a delay.

DracoLi (approved):

PR #467, opened 2025-03-13
Update pr-internal.yml

bpolania:
## Description
This PR fixes the issue with the unit test in the internal workflow

### DIFF
--- a/.github/workflows/pr-internal.yml
+++ b/.github/workflows/pr-internal.yml
@@ -1,4 +1,4 @@
-name: Conditional Workflow for External PRs with Unit & Integration Tests
+name: Conditional Workflow for Internal PRs with Unit & Integration Tests
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-internal.yml
```diff
@@ -24,7 +24,6 @@ jobs:
     secrets:
       WALLET_PRIVATE_KEY: ${{ secrets.WALLET_PRIVATE_KEY }}
       TEST_WALLET_ADDRESS: ${{ secrets.TEST_WALLET_ADDRESS }}
-      JUDGE_PRIVATE_KEY: ${{ secrets.JUDGE_PRIVATE_KEY }}
```
bonnie57, line 27: `push_build_and_test` also needs to remove `JUDGE_PRIVATE_KEY`. I am unsure if you are planning in the next pr.

This feature also needs to remove the variable from GitHub secrets.

bonnie57 (approved):
LGTM

.github/workflows/pr-internal.yml
bpolania: Yes, I need to change the reusable workflows first, will send a PR tomorrow

PR #468, opened 2025-03-13
Suffix with input about the related type

bonnie57:
## Description
- Add suffix with `DerivativeData`, `LicenseTermsData` and `LicensingConfig`


packages/core-sdk/src/types/common.ts
```diff
@@ -10,9 +10,9 @@ export type TypedData = {
 
 export type IpMetadataAndTxOptions = WithTxOptions & {
   /** The desired metadata for the newly minted NFT and newly registered IP. */
-  ipMetadata?: Partial<IpMetadataForWorkflow>;
+  ipMetadataInput?: Partial<IpMetadataForWorkflow>;
```
DracoLi, line 13: we don't need this change. we are only looking to rename type names, not input property names. this change will results in migrations for all users.

DracoLi (approved):

PR #471, opened 2025-03-14
Remove husky

DracoLi:
## Description

Removes husky from the codebase. 

I've had a number of issues with our husky hook that runs `pnpm fix`. It slows down commits and also has the side effect of adding all local changes to a commit which most of the time I do not want.

The main functionality of the `pnpm fix` right now is running lint fix and prettier. However, this does not guarantee all committed code is formatted because it is possible to get around hooks. We should instead be using CI to check for formatting and lint issues and its the PR owner's responsibility to resolve them before merging in code.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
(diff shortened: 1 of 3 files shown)


bonnie57 (approved):
LGTM

PR #472, opened 2025-03-14
Fix incorrect address/hex types

DracoLi:
## Description
 
1. Updated the `data` property in the following types from `Address` to `Hex`
    - `CancelDisputeRequest`, `IPAccountExecuteRequest`, `IPAccountExecuteWithSigRequest`, `ResolveDisputeRequest`
2. Change `mintFeeToken` and `owner` in `CreateNFTCollectionRequest` from `Hex` to `Address` type.

### DIFF
--- a/packages/core-sdk/src/types/resources/dispute.ts
+++ b/packages/core-sdk/src/types/resources/dispute.ts
@@ -42,7 +42,7 @@ export type RaiseDisputeResponse = {
 export type CancelDisputeRequest = {
   disputeId: number | string | bigint;
   /** Additional data used in the cancellation process. */
-  data?: Address;
+  data?: Hex;
   txOptions?: TxOptions;
 };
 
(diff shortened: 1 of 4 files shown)


packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -53,8 +53,11 @@ export type CancelDisputeResponse = {
 
 export type ResolveDisputeRequest = {
   disputeId: number | string | bigint;
-  /** The data to resolve the dispute. */
-  data: Address;
+  /**
+   * The data to resolve the dispute.
+   * Use `"0x"` if no additional data is needed.
+   */
+  data: Hex;
```
bonnie57, line 60: Do we need to set `0x` as the default value? It would be more user-friendly.

bonnie57 (approved):
LGTM

bonnie57:
I've reviewed all the code and identified a few issues, including:
- The `data` field in the `CancelDisputeRequest` type.
- The `data` field in the `IPAccountExecuteRequest` type.
- The `data` field in the `IPAccountExecuteWithSigRequest` type.

Could you add it?

packages/core-sdk/src/types/resources/dispute.ts
DracoLi: yes I agree, lets make that changes as a separate PR

PR #473, opened 2025-03-14
`AccountB` issue fix

bpolania:
## Description

This PR addresses the `AccountB` issue, which previously resulted in creating an unnecessary account and sending it 0.5 IP during tests that required two accounts. The earlier implementation relied on configuring two separate accounts within the `.env` file but was later modified to streamline the setup process for developers. However, the adopted method led to the creation and abandonment of multiple accounts containing small balances, cluttering the network unnecessarily.

The approach implemented in this PR uses `BIP32` to deterministically derive a extended private key (xpv) from the PK set in the `.env` file. It will fund the derived account if its balance goes under 5 IP and will set it up as `AccountB` for those tests requiring two accounts.

### DIFF
--- a/packages/core-sdk/test/integration/utils/BIP32.ts
+++ b/packages/core-sdk/test/integration/utils/BIP32.ts
@@ -30,9 +30,7 @@ export function getPrivateKeyFromXprv(xprv: string): Hex {
  */
 export function getXprvFromPrivateKey(privateKey: string | Hex): string {
   // Remove 0x prefix if present
-  const pkHex = privateKey.toString().startsWith("0x")
-    ? privateKey.toString().slice(2)
-    : privateKey.toString();
+  const pkHex = privateKey.toString().replace(/^0x/, "");
 
   // Validate private key length (must be exactly 32 bytes / 64 hex characters)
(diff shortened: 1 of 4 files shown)


packages/core-sdk/test/integration/utils/BIP32.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param privateKey Ethereum private key (with or without 0x prefix)
+ * @returns Deterministically derived extended private key (xprv)
+ */
+export function getXprvFromPrivateKey(privateKey: string | Hex): string {
+  // Remove 0x prefix if present
+  const pkHex = privateKey.toString().startsWith("0x")
+    ? privateKey.toString().slice(2)
+    : privateKey.toString();
+
```
bonnie57, line 36: It's minor, but using the following code will get better.
```
privateKey.toString().replace(/^0x/, "");
```

bonnie57 (approved):
LGTM

packages/core-sdk/test/integration/utils/BIP32.ts
bpolania: Done.

PR #474, opened 2025-03-17
Add collect and distribute group royalties

bonnie57:


packages/core-sdk/src/resources/group.ts
```diff
@@ -388,7 +395,78 @@ export class GroupClient {
...
+   * Emits an on-chain {@link https://github.com/storyprotocol/protocol-core-v1/blob/v1.3.1/contracts/interfaces/modules/grouping/IGroupingModule.sol#L38 | `CollectedRoyaltiesToGroupPool`} event.
+   */
+  public async collectAndDistributeGroupRoyalties({
+    groupIpId,
+    currencyTokens,
+    memberIpIds,
+    txOptions,
+  }: CollectAndDistributeGroupRoyaltiesRequest): Promise<CollectAndDistributeGroupRoyaltiesResponse> {
+    try {
+      const object: GroupingWorkflowsCollectRoyaltiesAndClaimRewardRequest = {
```
DracoLi, line 410: `object` is not a very good variable name. I recommend something like `protocolParams` or something similar to be a bit more descriptive on what this is for.

packages/core-sdk/src/resources/group.ts
```diff
@@ -388,7 +395,78 @@ export class GroupClient {
...
+      }
+      const isGroupRegistered = await this.ipAssetRegistryClient.isRegistered({
+        id: object.groupIpId,
+      });
+      if (!isGroupRegistered) {
+        throw new Error(`The group IP with ID ${object.groupIpId} is not registered.`);
+      }
+      await Promise.all(
+        object.memberIpIds.map(async (ipId) => {
+          const isMemberRegistered = await this.ipAssetRegistryClient.isRegistered({
```
DracoLi, line 432: Note: not needed to be addressed in this PR, just wanted to start a discussion.

I'm wondering if its actually worth checking for ip being registered? Each check will be an RPC call that slows things down and adds additional traffic to the RPC. I think its also very rare for clients to pass an invalid ip. Even if they do, the protocol call will just fail.
Thoughts? @bpolania @edisonz0718 @bonnie57

packages/core-sdk/src/types/resources/group.ts
```diff
@@ -105,3 +108,24 @@ export type RegisterGroupAndAttachLicenseAndAddIpsResponse = {
...
+
+export type CollectAndDistributeGroupRoyaltiesResponse = {
+  txHash: Hash;
+  receipts?: TransactionReceipt[];
+  collectedRoyalties?: Omit<GroupingModuleCollectedRoyaltiesToGroupPoolEvent, "pool">[];
+  royaltiesDistributed?: {
+    ipId: Address;
+    amount: bigint;
+    token: Address;
+    amountAfterFee: bigint;
```
DracoLi, line 129: can we add a jsdoc note on that this fee is for?

packages/core-sdk/test/integration/group.test.ts
```diff
@@ -255,4 +256,147 @@ describe("Group Functions", () => {
...
+      });
+      await client.royalty.payRoyaltyOnBehalf({
+        receiverIpId: childIpId2,
+        payerIpId: groupIpId,
+        token: WIP_TOKEN_ADDRESS,
+        amount: 100,
+        txOptions: { waitForTransaction: true },
+      });
+
+      //5. Transfer to vault
```
DracoLi, line 375: is this needed every-time before calling `collectAndDistributeGroupRoyalties` on the child ips? If so, do we need to expose an `transferToVault` method in the sdk as well?

packages/core-sdk/test/unit/resources/group.test.ts
```diff
@@ -553,4 +553,136 @@ describe("Test IpAssetClient", () => {
       expect(result.encodedTxData!.data).to.be.a("string").and.not.empty;
     });
   });
+
+  describe("Test groupClient.collectAndDistributeGroupRoyalties", async () => {
+    it("should throw error when call collectAndDistributeGroupRoyalties given group ip id is not registered", async () => {
```
DracoLi, line 558: lets try to keep the test name a bit more concise for improved readability. I feel like our test name are getting pretty long without conveying too much.
For example, I think this can be `it("throws error if group ipId is not registered")`

packages/core-sdk/src/resources/group.ts
edisonz0718: I think if protocol returns clear error message or error code, we can just use protocol's error return

packages/core-sdk/test/integration/group.test.ts
sebsadface: Yeah, it would be good to add a `transferToVault` method in the SDK if we haven't already. It doesn’t need to be called every time, only when there are royalties from non-direct descendants. If there are no claimable royalties in the royalty policy, the call to `transferToVault` will revert.

packages/core-sdk/src/resources/group.ts
DracoLi: @bonnie57 can you confirm if the protocol error is clear?

DracoLi (approved):

PR #477, opened 2025-03-18
fix: invalid default value for claim options

DracoLi:
## Description

Update jsdoc to fix the default value for ClaimAllRevenue options


bonnie57 (approved):
LGTM

PR #487, opened 2025-03-24
New `bond` limits fix

bpolania:
## Description

This PR updates the SDK to implement the new `bond` limits on `aeneid`


bonnie57 (approved):
minor thing, Could you use `parseEther` util method? It will look like more clear. 
Or extract the `minimumBond` variable.

PR #488, opened 2025-03-25
Fix issue about registerDerivative method

bonnie57:
## Description
Fix the wrong spender from `royaltyTokenDistributionWorkflows` to `royaltyModule` in the `registerDerivative` method.

### DIFF
--- a/packages/core-sdk/src/resources/ipAsset.ts
+++ b/packages/core-sdk/src/resources/ipAsset.ts
@@ -358,7 +358,7 @@ export class IPAssetClient {
           contractCall,
           txOptions: request.txOptions,
           encodedTxs: [encodedTxData],
-          spgSpenderAddress: this.royaltyTokenDistributionWorkflowsClient.address,
+          spgSpenderAddress: this.royaltyModuleEventClient.address,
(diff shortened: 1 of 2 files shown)


packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -358,7 +358,7 @@ export class IPAssetClient {
           contractCall,
           txOptions: request.txOptions,
           encodedTxs: [encodedTxData],
-          spgSpenderAddress: this.royaltyTokenDistributionWorkflowsClient.address,
+          spgSpenderAddress: this.royaltyModuleEventClient.address,
```
DracoLi, line 361: is this still wrong, it should be `licensingModuleClient`.

bonnie57: It should be `royaltyModule`.    Because we should approve the royalty module can transfer some token from wallet to another address.

DracoLi: yes you are right, it should be the royalty module.

DracoLi (approved):

PR #491, opened 2025-03-25
Improve document

bonnie57:
## Description
- Modify the `disputeAssertion` method description
- Add the `chainId` inline doc.


packages/core-sdk/src/types/config.ts
```diff
@@ -28,6 +28,11 @@ export type UseWalletStoryConfig = {
 
 export type StoryConfig = {
   readonly transport: Transport;
+  /**
+   * The chain ID to use.
```
DracoLi, line 32: lets mention that this defaults to the aeneid change.

PR #494, opened 2025-04-01
Add minimum bond in raiseDispute

bonnie57:
## Description
- Add minimum bond check
- Make `bond` optional and use minimum bond as default value.

### DIFF
--- a/packages/core-sdk/src/resources/dispute.ts
+++ b/packages/core-sdk/src/resources/dispute.ts
@@ -56,25 +56,22 @@ export class DisputeClient {
   public async raiseDispute(request: RaiseDisputeRequest): Promise<RaiseDisputeResponse> {
     try {
       const liveness = BigInt(request.liveness);
-      const bonds = BigInt(request.bond);
       const [minLiveness, maxLiveness] = await Promise.all([
         this.arbitrationPolicyUmaClient.minLiveness(),
         this.arbitrationPolicyUmaClient.maxLiveness(),
       ]);
 
+      const [minimumBond, maxBonds] = await Promise.all([
+        getMinimumBond(this.rpcClient, this.arbitrationPolicyUmaClient, WIP_TOKEN_ADDRESS),
+        this.arbitrationPolicyUmaClient.maxBonds({
+          token: WIP_TOKEN_ADDRESS,
+        }),
+      ]);
+      const bonds = BigInt(request.bond || minimumBond);
       const tag = stringToHex(request.targetTag, { size: 32 });
       if (liveness < minLiveness || liveness > maxLiveness) {
         throw new Error(`Liveness must be between ${minLiveness} and ${maxLiveness}.`);
       }
-
-      const maxBonds = await this.arbitrationPolicyUmaClient.maxBonds({
-        token: WIP_TOKEN_ADDRESS,
-      });
-      const minimumBond = await getMinimumBond(
-        this.rpcClient,
-        this.arbitrationPolicyUmaClient,
-        WIP_TOKEN_ADDRESS,
-      );
       if (bonds > maxBonds || bonds < minimumBond) {
         throw new Error(`Bonds must be between ${minimumBond} and ${maxBonds}.`);
       }
--- a/packages/core-sdk/src/resources/dispute.ts
+++ b/packages/core-sdk/src/resources/dispute.ts
@@ -60,21 +60,20 @@ export class DisputeClient {
         this.arbitrationPolicyUmaClient.minLiveness(),
         this.arbitrationPolicyUmaClient.maxLiveness(),
       ]);
-
-      const [minimumBond, maxBonds] = await Promise.all([
+      if (liveness < minLiveness || liveness > maxLiveness) {
+        throw new Error(`Liveness must be between ${minLiveness} and ${maxLiveness}.`);
+      }
+      const [minimumBond, maximumBond] = await Promise.all([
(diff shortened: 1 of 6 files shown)


packages/core-sdk/test/unit/resources/dispute.test.ts
```diff
@@ -122,7 +141,7 @@ describe("Test DisputeClient", () => {
         targetIpId: "0x1daAE3197Bc469Cb97B917aa460a12dD95c6627c",
         cid: "QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR",
         targetTag: "tag",
-        bond: 0,
+        bond: 15,
```
bpolania, line 144: Can we make this value smaller? so we don't drain the wallet too fast? If not is OK

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -56,22 +56,24 @@ export class DisputeClient {
   public async raiseDispute(request: RaiseDisputeRequest): Promise<RaiseDisputeResponse> {
     try {
       const liveness = BigInt(request.liveness);
-      const bonds = BigInt(request.bond);
       const [minLiveness, maxLiveness] = await Promise.all([
         this.arbitrationPolicyUmaClient.minLiveness(),
         this.arbitrationPolicyUmaClient.maxLiveness(),
       ]);
 
+      const [minimumBond, maxBonds] = await Promise.all([
```
DracoLi, line 64: minor, but we should do this after the liveness validation. we don't need 2 rpc queries to fetch bonds if other validation will fail.
In terms of priority, 1st validate all inputs that don't require rpc calls, then validate each input that require rpc call one after another.

packages/core-sdk/src/types/resources/dispute.ts
```diff
@@ -25,7 +25,7 @@ export type RaiseDisputeRequest = WithTxOptions & {
    * To counter that dispute the opposite party of the dispute has to place a bond of the same amount.
    * The winner of the dispute gets the original bond back + 50% of the other party bond. The remaining 50% of the loser party bond goes to the reviewer.
    */
-  bond: bigint | number | string;
+  bond?: bigint | number | string;
```
DracoLi, line 28: add a note on the default value for bonds

packages/core-sdk/src/utils/oov3.ts
```diff
@@ -23,3 +23,18 @@ export const getAssertionDetails = async (
 
   return bond;
 };
+
+export const getMinimumBond = async (
```
bpolania, line 27: How are you handling the error if the contract call fails? perhaps use 'try/catch'?

packages/core-sdk/test/unit/resources/dispute.test.ts
```diff
@@ -31,6 +31,9 @@ describe("Test DisputeClient", () => {
   });
 
   describe("raiseDispute", () => {
+    beforeEach(() => {
+      rpcMock.readContract = sinon.stub().resolves(10n);
```
DracoLi, line 35: is not clear this is for mocking the min bonds value. please consider adding a comment and also make the `10n` a variable like `mockMinBonds`. Then in the following code, you can do `mockMinBonds + 5` instead of a random number, which will make it clearer than you are doing.

packages/core-sdk/test/unit/resources/dispute.test.ts
DracoLi: not needed, this is unit test

packages/core-sdk/src/utils/oov3.ts
DracoLi: this should not be needed, the caller should be handling errors. I think its currently handled well.

packages/core-sdk/src/utils/oov3.ts
```diff
@@ -23,3 +23,18 @@ export const getAssertionDetails = async (
...
+  rpcClient: PublicClient,
+  arbitrationPolicyUmaClient: ArbitrationPolicyUmaClient,
+  currency: Address,
+) => {
+  const oov3Contract = await getOov3Contract(arbitrationPolicyUmaClient);
+  const minimumBond = await rpcClient.readContract({
```
bpolania, line 33: Since you are returning `minimumBond`right away, maybe just use:
```
return await rpcClient.readContract({
    address: oov3Contract,
    abi: ASSERTION_ABI,
    functionName: "getMinimumBond",
    args: [currency],
  });
```
 This is just a suggestion but feel free to ignore.

packages/core-sdk/test/unit/resources/dispute.test.ts
bonnie57: It's very clear.

packages/core-sdk/src/resources/dispute.ts
```diff
@@ -56,22 +56,24 @@ export class DisputeClient {
...
       if (liveness < minLiveness || liveness > maxLiveness) {
         throw new Error(`Liveness must be between ${minLiveness} and ${maxLiveness}.`);
       }
-
-      const maxBonds = await this.arbitrationPolicyUmaClient.maxBonds({
-        token: WIP_TOKEN_ADDRESS,
-      });
-      if (bonds > maxBonds) {
-        throw new Error(`Bonds must be less than ${maxBonds}.`);
+      if (bonds > maxBonds || bonds < minimumBond) {
```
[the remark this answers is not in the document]
bonnie57: Agree. I have changed the name of `maximumBond`. It's better than `maxBond`.

DracoLi (approved):

PR #498, opened 2025-04-02
Origin/bp/flow working

bpolania:
## Description
Summary of Changes in this PR:

* Removed dev as a target branch from existing workflows.
* Eliminated all references to the React-SDK.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -43,11 +43,8 @@ jobs:
       - name: Set Configuration File
         id: set_config
         run: |
-          if [ "${{ inputs.is_core_sdk }}" = "true" ]; then
-            echo "config_file=configs/configuration.json" >> $GITHUB_ENV
-          else
-            echo "config_file=configs/configuration_react.json" >> $GITHUB_ENV
-          fi
+          echo "config_file=configs/configuration.json" >> $GITHUB_ENV
+         
 
       - name: "Build Changelog"
         id: github_release
--- a/.github/workflows/pr-external.yaml
+++ b/.github/workflows/pr-external.yaml
@@ -4,7 +4,7 @@ on:
   pull_request_target:
     types: [opened, synchronize]
     branches:
-      - dev
+      - main
 
 jobs:
   Timestamp_PR_CREATED:
--- a/.github/workflows/publish-package.yaml
+++ b/.github/workflows/publish-package.yaml
@@ -1,13 +1,11 @@
 name: Publish to npm, Tag and create GH Release
 
 on:
-  push:
-    branches:
-      - main
-
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    # This allows manual triggering of the workflow
 
 permissions:
   contents: write
@@ -22,7 +20,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       CORE_SDK_VERSION_TO_BE_PUBLISHED: ${{ steps.get_version_to_publish.outputs.CORE_SDK_VERSION_TO_BE_PUBLISHED }}
-      REACT_SDK_VERSION_TO_BE_PUBLISHED: ${{ steps.get_version_to_publish.outputs.REACT_SDK_VERSION_TO_BE_PUBLISHED }}
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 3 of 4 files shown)


bonnie57:
Before we remove the `dev` branch, I think we need to reconsider the PR workflow for integration tests. The external workflow should include the integration test. Both the external and internal workflows need to ensure that the integration test is good before it can be merged.

.github/workflows/create-release.yml
bonnie57, line 1: Need to refactor the workflow variables. Because we don't need to distinguish `react-sdk` and `core-sdk` in this workflow. 
- `tag_name` should be renamed to `tag_version` to refer to the specific publish version.
- `is_core_sdk` should be removed.
- `last_tag` can be removed and get in this workflow.

.github/workflows/create-release.yml
```diff
@@ -43,11 +43,8 @@ jobs:
       - name: Set Configuration File
         id: set_config
         run: |
-          if [ "${{ inputs.is_core_sdk }}" = "true" ]; then
-            echo "config_file=configs/configuration.json" >> $GITHUB_ENV
-          else
-            echo "config_file=configs/configuration_react.json" >> $GITHUB_ENV
-          fi
+          echo "config_file=configs/configuration.json" >> $GITHUB_ENV
+
```
bonnie57, line 47: Need to delete the `configs/configuration_react.json` file.

.github/workflows/create-release.yml
bonnie57, line 1: I remember the previous decision was that we needed to manually trigger the tag, and the workflow needs to be triggered and publish a new package.  If we follow the decision, `tag_name` should be removed, and `Tag and Push` steps also need be removed.

.github/workflows/pr-external.yaml
```diff
@@ -30,4 +30,4 @@ jobs:
       ENVIRONMENT: "odyssey"
```
bonnie57, line 30: `odyssey` should rename to `test`. So we don't change again.

.github/workflows/publish-package.yaml
```diff
@@ -169,52 +131,21 @@ jobs:
...
-      tag_name: "@story-protocol/react-sdk@${{ needs.print_version_to_publish.outputs.react_sdk_version_to_be_published }}"
-      is_core_sdk: false
-      last_tag: ${{ needs.fetch_last_tag.outputs.REACT_SDK_LATEST_TAG }}
-
-  send_slack_notif-react-sdk:
-    needs: [print_version_to_publish, create-release-react-sdk]
-    uses: storyprotocol/gha-workflows/.github/workflows/reusable-slack-notifs.yml@main
-    with:
-      short-desc: "${{ github.repository }}: @story-protocol/react-sdk package has been published to NPM Registry, version: ${{ needs.print_version_to_publish.outputs.react_sdk_version_to_be_published }}"
+      short-desc: "${{ github.repository }}: @story-protocol/core-sdk package has been published to NPM Registry, version: ${{ needs.print_version_to_publish.outputs.CORE_SDK_VERSION_TO_BE_PUBLISHED}}"
```
bonnie57, line 145: Could you specify `channel-name` of `sdk-channel`? Because it's important information for `sdk-channel`.

.github/workflows/publish-package.yaml
```diff
@@ -1,13 +1,11 @@
...
 on:
-  push:
-    branches:
-      - main
-
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    # This allows manual triggering of the workflow
```
DracoLi, line 8: i think this can stay like this for now. but we should make sure to follow up with the ability to auto publish on new tags if the sdk version is newer.

.github/workflows/create-release.yml
bpolania: Yes, the rest of the react related changes, including this one, is being handled in PR #499

.github/workflows/pr-external.yaml
bpolania: We'll update to `aeneied`

.github/workflows/create-release.yml
bpolania: Updated

bpolania: `tag_name` removed, will confirm with @DracoLi about manual tagging

.github/workflows/publish-package.yaml
bpolania: Not sure I understand this one. You want to have the channel name in the `short-desc`?

DracoLi (approved):

PR #504, opened 2025-04-07
Add addIp method 

bonnie57:
## Description  
This PR implements the `addIpsToGroup` method in the group module, mirroring the functionality of the [Smart Contract `addIp` method](https://github.com/storyprotocol/protocol-core-v1/blob/main/contracts/modules/grouping/GroupingModule.sol#L149).

## Changes  
- Added `addIpsToGroup` method to the `group` module  
- Introduced corresponding `AddIpRequest` type in `group` types  

## Testing  
**Unit Tests Added:**  
- Validates error throwing for invalid IP addresses  
- Verifies successful transaction hash return  

**Integration Test Added:**  
- Ensures successful addition of multiple IPs to a group

### DIFF
--- a/packages/core-sdk/src/types/resources/group.ts
+++ b/packages/core-sdk/src/types/resources/group.ts
@@ -148,7 +148,8 @@ export type AddIpRequest = {
   /**
    * The maximum reward share percentage that can be allocated to each member IP.
    * Must be between 0 and 100 (where 100% represents 100_000_000).
+   * @default 100
    */
-  maxAllowedRewardShare: number | string;
+  maxAllowedRewardSharePercentage?: number;
   txOptions?: Omit<TxOptions, "encodedTxDataOnly">;
 };
--- a/packages/core-sdk/src/types/resources/group.ts
+++ b/packages/core-sdk/src/types/resources/group.ts
@@ -137,3 +137,18 @@ export type CollectAndDistributeGroupRoyaltiesResponse = {
     amountAfterFee: bigint;
   }[];
 };
+
+export type AddIpRequest = {
+  groupIpId: Address;
+  /**
+   * The addresses of the IPs to add to the Group IP.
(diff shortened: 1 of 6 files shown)


packages/core-sdk/src/types/resources/group.ts
```diff
@@ -137,3 +137,18 @@ export type CollectAndDistributeGroupRoyaltiesResponse = {
...
+  /**
+   * The addresses of the IPs to add to the Group IP.
+   * IP IDs must be attached to the group IP license terms.
+   */
+  ipIds: Address[];
+  /**
+   * The maximum reward share percentage that can be allocated to each member IP.
+   * Must be between 0 and 100 (where 100% represents 100_000_000).
+   */
+  maxAllowedRewardShare: number | string;
```
DracoLi, line 152: lets avoid support `string` for numbers

packages/core-sdk/src/types/resources/group.ts
```diff
@@ -137,3 +137,18 @@ export type CollectAndDistributeGroupRoyaltiesResponse = {
...
+export type AddIpRequest = {
+  groupIpId: Address;
+  /**
+   * The addresses of the IPs to add to the Group IP.
+   * IP IDs must be attached to the group IP license terms.
+   */
+  ipIds: Address[];
+  /**
+   * The maximum reward share percentage that can be allocated to each member IP.
+   * Must be between 0 and 100 (where 100% represents 100_000_000).
```
DracoLi, line 150: does it make sense to have a default for this? maybe 100%?

packages/core-sdk/src/types/resources/group.ts
DracoLi: since its a percentage, do you think its clearer if we name this `maxAllowedRewardSharePercentage`?

bpolania (approved):

DracoLi (approved):

PR #505, opened 2025-04-10
Add batchRegisterIpAssetsWithOptimizedWorkflows method

bonnie57:
## Add batchRegisterIpAssetsWithOptimizedWorkflows method

The batch register should contain the following methods:
- mintAndRegisterIpAndMakeDerivative

- mintAndRegisterIpAssetWithPilTerms

- mintAndRegisterIpAndAttachPilTermsAndDistributeRoyaltyTokens

- mintAndRegisterIpAndMakeDerivativeAndDistributeRoyaltyTokens

- registerDerivativeIpAndAttachLicenseTermsAndDistributeRoyaltyTokens

- registerIpAndAttachPilTerms

- registerIPAndAttachLicenseTermsAndDistributeRoyaltyTokens
 
- registerDerivativeIp


packages/core-sdk/src/types/options.ts
```diff
@@ -69,13 +69,13 @@ export type WithWipOptions = {
   wipOptions?: WipOptions;
 };
 export type HandleTxOptionsParams = {
-  txHash: Hex;
+  txHash: Hash | Hash[];
```
DracoLi, line 72: instead of this, i think we can do some clean up here. `handleTxOptions` is not too clear on what the method does. How about renaming this to `waitForTxReceipt` and `waitForTxReceipts` this one takes in `txHashes`. Logic for `waitForTxReceipt` should just call `waitForTxReceipts`

packages/core-sdk/src/utils/feeUtils.ts
```diff
@@ -203,7 +203,7 @@ export const contractCallWithFees = async ({
   encodedTxs,
   rpcClient,
   token,
-}: ContractCallWithFees): Promise<TransactionResponse> => {
+}: ContractCallWithFees): Promise<TransactionResponse | TransactionResponse[]> => {
```
DracoLi, line 206: making this a join makes the typing much harder to work with. I recommend either creating separate version that returns or make `TransactionResponse[]` the only returned type.

packages/core-sdk/src/types/resources/permission.ts
```diff
@@ -83,7 +83,7 @@ export type PermissionSignatureRequest = {
   state: Hex;
   /** The deadline for the signature in seconds. */
   deadline: string | number | bigint;
-  wallet: WalletClient;
+  wallet: SimpleWalletClient;
```
DracoLi, line 86: why was this change needed?

packages/core-sdk/src/resources/royalty.ts
```diff
@@ -387,7 +389,7 @@ export class RoyaltyClient {
       txHash,
       rpcClient: this.rpcClient,
       txOptions,
-    });
+    }) as Promise<TransactionResponse>;
```
DracoLi, line 392: we need to minimize the amount of casing in the code. Lets make those method updates so we can avoid doing it, thus making the code cleaner.

packages/core-sdk/src/types/resources/permission.ts
bonnie57: We totally use the `SimpleWalletClient `type in modules. If we use `WalletClient `, it will cause a type conflict.

packages/core-sdk/src/types/options.ts
bonnie57: Good Point. 👍

packages/core-sdk/src/utils/feeUtils.ts
bonnie57: I have a solution to solve the issue. I added a `generic` type in the `contractCallWithFees`, and the default type is `Hash`, which is a common case.  A separate version is difficult to maintain. Because there are much logic in the method. What do you think?

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -310,7 +326,7 @@ export type RegisterIPAndAttachLicenseTermsAndDistributeRoyaltyTokensResponse =
 };
 export type DistributeRoyaltyTokens = {
   ipId: Address;
-  deadline: bigint;
+  deadline?: bigint | string | number;
```
bonnie57, line 329: It's hard to remove `string`. `getCalculatedDead` moves to `registerHelper` file. I need to pass the primitive type to this method.

We can create a ticket to remove string type from number type.

DracoLi: sounds good. we can clean these up in 1.4.0. For this release, everything still needs to be backwards compatible.

packages/core-sdk/src/types/utils/wip.ts
```diff
@@ -42,12 +48,12 @@ export type TokenApprovalCall = {
...
 };
 
-export type ContractCallWithFees = {
+export type ContractCallWithFees<T extends Hash | Hash[] = Hash> = {
   totalFees: bigint;
   multicall3Address: Address;
   /** all possible spenders of the erc20 token */
   tokenSpenders: Erc20Spender[];
-  contractCall: () => Promise<Hash>;
+  contractCall: () => Promise<T>;
```
DracoLi, line 56: instead of supporting both? can we just make this return Hash[] by default? I added complexity doesn't seem worth it for me versus just returning an array for all instead. Same for `MulticallWithWrapIp`

packages/core-sdk/src/utils/feeUtils.ts
```diff
@@ -276,5 +277,20 @@ export const contractCallWithFees = async ({
     wallet,
     calls,
   });
-  return handleTxOptions({ rpcClient, txOptions, txHash });
+  return handleTransactionResponse(txHash, rpcClient, txOptions) as ContractCallWithFeesResponse<T>;
```
DracoLi, line 280: all these forced casting need to be removed. Lets just change this to only support `Hash[]`. I feel like that is the simpliest approach.

packages/core-sdk/src/utils/feeUtils.ts
```diff
@@ -276,5 +277,20 @@ export const contractCallWithFees = async ({
...
+};
+
+export const handleTransactionResponse = <T extends Hash | Hash[] = Hash>(
```
DracoLi, line 283: lets do both `handleTransactionResponse` and `handleTransactionsResponse`. It won't be too much extra code since `handleTransactionResponse` can just call `handleTransactionsResponse`. I think it will make the code more maintainable.

packages/core-sdk/src/utils/registerHelper.ts
```diff
@@ -0,0 +1,1319 @@
...
+    ipId,
+    royaltyShares: request.royaltyShares,
+    sigApproveRoyaltyTokens: {
+      signer: wallet.account!.address,
+      deadline: calculatedDeadline,
+      signature: signatureApproveRoyaltyTokens,
+    },
+  };
+  return {
+    transformRequest: transformRequest as T,
```
DracoLi, line 358: casting really defeat the purchase of all the typing work that we are doing. We should avoid casting as much as possible but some casts are necessary so I suggest adding a comment above every cast used indicating why its absolutely needed.

packages/core-sdk/src/utils/registerHelper.ts
```diff
@@ -0,0 +1,1319 @@
...
+    isUseMulticall3: false,
```
DracoLi, line 359: just to confirm: this workflow doesn't use multicall anyways even if this is enabled right?

packages/core-sdk/src/utils/registerHelper.ts
```diff
@@ -0,0 +1,1319 @@
...
+      throw new Error("The sum of the royalty shares cannot exceeds 100.");
+    }
+    const value = (share.percentage / 100) * royaltySharesTotalSupply;
+    actualTotal += value;
+    return { ...share, percentage: value };
+  });
+  return { royaltyShares: shares, totalAmount: actualTotal } as const;
+};
+
+export const validateDerivativeData = async ({
```
DracoLi, line 515: this file is getting pretty big (+1k lines), maybe we move the validation helpers into a separate file? I think there are also other related utils in this file that we can group together and make separate files for.

packages/core-sdk/src/utils/registerHelper.ts
```diff
@@ -0,0 +1,1319 @@
...
+  } else if ("nftContract" in request && "tokenId" in request) {
+    return handleRegisterRequest<T>({ request, rpcClient, wallet, chainId });
+  } else if ("ipRoyaltyVault" in request) {
+    return handleDistributeRoyaltyTokensRequest<T>({ request, rpcClient, wallet, chainId });
+  }
+
+  throw new Error("Invalid registration request type");
+};
+
+const aggregateTransformIpRegistrationWorkflow = (
```
DracoLi, line 1233: lets add some comments to some of these helper functions. I think its not clear from the name exactly what their responsibilities are.

packages/core-sdk/src/types/resources/ipAsset.ts
bonnie57: Created it.

packages/core-sdk/src/types/resources/ipAsset.ts
```diff
@@ -422,3 +438,80 @@ export type RegistrationResponse = {
...
+  | MintAndRegisterIpAndMakeDerivativeRequest
+  | MintAndRegisterIpAssetWithPilTermsRequest
+  | MintAndRegisterIpAndAttachPILTermsAndDistributeRoyaltyTokensRequest
+  | MintAndRegisterIpAndMakeDerivativeAndDistributeRoyaltyTokensRequest
+>;
+
+export type IpRegistrationWorkflowRequest =
+  | MintSpgNftRegistrationRequest
+  | RegisterRegistrationRequest
+  | RemoveOptionsFields<DistributeRoyaltyTokens>;
```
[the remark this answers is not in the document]
bonnie57: Good point.

packages/core-sdk/src/utils/registerHelper.ts
bonnie57: I'm sorry the code is outdated. Do you mean `mintAndRegisterIpAndAttachPilTermsAndDistributeRoyaltyTokens` method?
The smart contract doesn't support.  Once the contract engineer changes, we can support it.

packages/core-sdk/src/types/utils/wip.ts
bonnie57: If we accept `Hash[]`, `contractCallWithFees` return type is `Promise<TransactionResponse[]>`. We also face a similar issue like `handleTxOptions`. We need to judge whether the array is or not when calling `contractCallWithFees`. SDK has many places that call the `contractCallWithFees` method.

bpolania (approved):
LGTM

DracoLi (approved):

PR #508, opened 2025-04-23
Add support for public minting checks in IP asset registration

bonnie57:
## Description
    
If the SPG NFT contract's public minting is disabled, we need to check if the caller has the `minter role`.
When public minting is disabled, we can't use `multicall` because we need to perform additional role checks
that aren't compatible with batched transactions.

This is because role-based access control requires the transaction's `msg.sender` to be verified directly,
which is not preserved when using `multicall` (where the multicall contract becomes the sender).

The following methods will be checked:
- mintAndRegisterIpAssetWithPilTerms
- mintAndRegisterIpAndMakeDerivative
- mintAndRegisterIp
- mintAndRegisterIpAndMakeDerivativeWithLicenseTokens
- mintAndRegisterIpAndAttachPilTermsAndDistributeRoyaltyTokens
- mintAndRegisterIpAndMakeDerivativeAndDistributeRoyaltyTokens
      

## Test Plan 
Add the integration test for private mint SPG NFT contract.

### DIFF
--- a/packages/core-sdk/src/resources/ipAsset.ts
+++ b/packages/core-sdk/src/resources/ipAsset.ts
@@ -106,6 +106,7 @@ import { handleMulticall } from "../utils/registrationUtils/registerHelper";
 import {
   getCalculatedDeadline,
   getIpIdAddress,
+  getPublicMinting,
 } from "../utils/registrationUtils/registerValidation";
 import {
   getRoyaltyShares,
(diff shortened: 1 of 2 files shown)


packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -1530,12 +1531,25 @@ export class IPAssetClient {
...
+    let useMulticallWhenPossible = wipOptions?.useMulticallWhenPossible ?? true;
 
     // get spg minting fee
     if (spgNftContract) {
       const nftMintFee = await calculateSPGWipMintFee(
         new SpgnftImplReadOnlyClient(this.rpcClient, spgNftContract),
       );
+      const publicMinting = await getPublicMinting(spgNftContract, this.rpcClient);
+      /**
+       * If the SPG NFT contract's public minting is disabled, we need to check if the caller has the `minter role`.
```
DracoLi, line 1543: awesome, thx for adding this :+1:

DracoLi (approved):

packages/core-sdk/test/integration/ipAsset.test.ts
```diff
@@ -1261,6 +1261,190 @@ describe("IP Asset Functions", () => {
...
       });
     });
+
+    describe("IP Asset Registration with Private Minting SPG NFT Contracts", () => {
+      let spgNftContractWithPrivateMinting: Address;
+      let licenseTermsId: bigint;
+      let parentIpId: Address;
+      beforeEach(async () => {
+        const privateMintingCollectionResult = await client.nftClient.createNFTCollection({
+          name: "Public Minting Collection",
```
bpolania, line 1271: Not a big thing, but if this test is for private minting  and the `isPublicMinting` is set to false, perhaps the name of the NFT Collection shouldn't be "Public Minting Collection"

bpolania (dismissed):
Left a small comment, but in general LGTM

packages/core-sdk/test/integration/ipAsset.test.ts
bonnie57: Good points.

bpolania (approved):

DracoLi (approved):

PR #509, opened 2025-04-23
IPAsset - Batch Integration Tests

bpolania:
## Description
This PR adds integration tests for batch register IP functionality


packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+  erc20Address,
+} from "../../src/abi/generated";
+import { WIP_TOKEN_ADDRESS } from "../../src/constants/common";
+import { ERC20Client } from "../../src/utils/token";
+
+chai.use(chaiAsPromised);
+const expect = chai.expect;
+
+const pool = evenSplitGroupPoolAddress[aeneid];
+const walletAddress = process.env.TEST_WALLET_ADDRESS! as Address;
```
bonnie57, line 25: Minor thing. In the `integration/util/util` file, there is a variable called `TEST_WALLET_ADDRESS`. You can directly import it.

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+    });
+    nftContract = txData.spgNftContract!;
+    
+
+    mockERC20 = new ERC20Client(
+      publicClient,
+      walletClient,
+      erc20Address[aeneid]
+    );
+    await mockERC20.mint(walletAddress, 100000n);
```
bonnie57, line 53: It's the same. We need to consider the `WIP` token instead of `ERC20`, which is the common case.

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+        expect(result.ipRoyaltyVault).to.be.a("string").and.not.empty;
+        
+        const isRegistered = await client.ipAsset.isRegistered(result.ipId!);
+        expect(isRegistered).to.be.true;
+      }
+
+      const batchRequest = {
+        requests: requests.map(req => ({
+          ...req,
+        })),
```
bonnie57, line 163: You don't need to map it. You can write it as follows:
```
 const batchRequest = {
        requests,
      };
```

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+      };
+
+      const batchResult = await client.ipAsset.batchRegisterIpAssetsWithOptimizedWorkflows(batchRequest);
+      expect(batchResult.registrationResults).to.be.an("array");
+      expect(batchResult.registrationResults.length).to.be.lessThan(requests.length);
+
```
bonnie57, line 169: I think we need to provide an exact number instead of a rough estimate. It will provide exact feedback.

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+      const maxRecipientsPercentage = 10; 
+      const maxRecipients = 10; 
+      
+      const maxRecipientsDistribution = Array(maxRecipients).fill(0).map(() => ({
+        recipient: walletAddress,
+        percentage: maxRecipientsPercentage,
+      }));
+
+      const totalPercentage = maxRecipientsDistribution.reduce((sum, item) => sum + item.percentage, 0);
+      expect(totalPercentage).to.equal(100);
```
bonnie57, line 210: It's your own logic, we don't need to assert.

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+      }
+      
+      expect(ipIds.length).to.equal(batchRequest.requests.length);
+      for (const ipId of ipIds) {
+        const isRegistered = await client.ipAsset.isRegistered(ipId as `0x${string}`);
+        expect(isRegistered).to.be.true;
+      }
+    });
+  });
+});
```
bonnie57, line 326: How to verify royalty token ownership? I don't see the related assertion.

packages/core-sdk/test/integration/ipAsset.batch.test.ts
```diff
@@ -0,0 +1,715 @@
...
+
+chai.use(chaiAsPromised);
+const expect = chai.expect;
+
+const pool = evenSplitGroupPoolAddress[aeneid];
+const walletAddress = process.env.TEST_WALLET_ADDRESS! as Address;
+
+describe("Batch IP Registration with Royalty Distribution", () => {
+  let client: StoryClient;
+  let nftContract: Hex;
```
bonnie57, line 29: Could you rename the `spgNftContract`? Because in the IPAsset module, there are two kinds of NFT contracts, namely SPG NFT contract and NFT contract.

PR #526, opened 2025-05-06
Return claimedTokens to claimAllRevenue when not owner

bonnie57:


packages/core-sdk/src/resources/royalty.ts
```diff
@@ -118,7 +113,7 @@ export class RoyaltyClient {
         });
         txHashes.push(...hashes);
       }
-      if (autoUnwrapIp) {
+      if (autoUnwrapIp && ownsClaimer) {
```
bpolania, line 116: Looking at the `claimAllRevenue` function, I noticed that it only reports the `Failed to claim all revenue` error that will trigger if the `royaltyWorkflowsClient.claimAllRevenue` call fails. However, two other calls for transfer and unwrap tokens may fail even if the revenue claim succeeds. 

Now, `transferClaimedTokensFromIpToWallet` doesn't catch any errors, and `unwrapWipTokens` only throws an error if `Multiple WIP tokens found in the claimed tokens.`

So, if the claim succeeds but the transfer or the unwrap fails the user might not get any error.

bonnie57: We don't need to catch errors separately, it is always watched by the `claimAllRevenue` method.

bpolania: My point is that if the claim can succeed, the transfer or the wrapping might fail. So, giving a "Failed to claim all revenue" might be misleading, right?

bonnie57: It will always show the error stack . If the `transferClaimedTokensFromIpToWallet` catches the error, it will not show to users.

bpolania (approved):

DracoLi (approved):

lucas2brh (approved):

PR #538, opened 2025-05-15
Update ESLint configuration to enforce alphabetical ordering of name

bonnie57:
## Description
We already have `import/order`, so we no longer need to add [sort-imports](https://eslint.org/docs/latest/rules/sort-imports#rule-details).  

Update ESLint configuration to enforce alphabetical ordering of names to keep `import order` consistent.


bpolania (approved):
LGTM

lucas2brh (approved):

PR #540, opened 2025-05-15
Test Workflow Improvements

bpolania:
## Description
This PR fixes the current unit/integration tests workflow. 

* Sets integration tests only after a PR is submitted.
* Sets unit tests only after a push. 
* Fixes the issues with the notification to 57Blocks slack channel

### DIFF
--- a/.github/workflows/publish-package.yaml
+++ b/.github/workflows/publish-package.yaml
@@ -119,3 +119,15 @@ jobs:
           npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  
+  send_slack_notif-core-sdk:
+    needs: [print_version_to_publish]
+    uses: storyprotocol/gha-workflows/.github/workflows/reusable-slack-notifs.yml@main
+    with:
+      short-desc: "${{ github.repository }}: @story-protocol/core-sdk package has been published to NPM Registry, version: ${{ needs.print_version_to_publish.outputs.CORE_SDK_VERSION_TO_BE_PUBLISHED}}"
+      title: "Published to Registry"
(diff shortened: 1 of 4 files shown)


bonnie57 (commented):
We need to trigger `pr-internal` and `pr-external` when a pull request is created. 
Then we need to run unit tests and integration tests when pr code changes.

.github/workflows/publish-package.yaml
```diff
@@ -118,16 +118,4 @@ jobs:
           cd packages/core-sdk
           npm publish
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-
-  send_slack_notif-core-sdk:
```
bonnie57, line 123: We cannot directly remove it. We should trigger when a new package is released.  Besides, we need to update the channel to the SDK channel.

bpolania: Since this is a manual process now, I'm not sure if it's worth having that notification in that channel.

.github/workflows/publish-package.yaml
```diff
@@ -119,15 +119,17 @@ jobs:
...
\ No newline at end of file
+    runs-on: ubuntu-latest
+    environment: "odyssey"
+    steps:
+      - name: Send Slack Notification
+        run: |
+          curl -X POST \
+            -H 'Content-type: application/json' \
+            --data '{
+              "channel": "qa-messenger",
```
AndyBoWu, line 132: you might wanna update the channel name too, @bpolania

bpolania: Fixed

bonnie57 (approved):
LGTM

AndyBoWu (approved):
LGTM

PR #545, opened 2025-05-22
Fix txHash type

bonnie57:
## Description
Fix `txHash` type from `string` and `Hex` into `txHash`.

## Notes
Removing `txHash` when attachLicenseTerms given licenseTermsId is already attached. @bpolania, there is a breaking change, and we need to highlight it in the release notes.


lucas2brh (approved):

bpolania (approved):
LGTM

Issue #579, opened 2025-07-16
Expose Flavor Structs


lucas2brh:
[https://www.notion.so/storyprotocol/Expose-Flavor-Structs-1ea051299a548028bd95e9a49e0b074e?source=copy_link](https://www.notion.so/storyprotocol/Expose-Flavor-Structs-1ea051299a548028bd95e9a49e0b074e?source=copy_link)

Issue #576, opened 2025-07-15
PILFlavor


lucas2brh:
[https://www.notion.so/storyprotocol/TS-SDK-v1-4-0-rc-1-Planned-Changes-234051299a54812589c5ee659c17072f?source=copy_link](https://www.notion.so/storyprotocol/TS-SDK-v1-4-0-rc-1-Planned-Changes-234051299a54812589c5ee659c17072f?source=copy_link)

PR #605, opened 2025-07-25
Refactor PIL Flavor

bonnie57:
## Description
- Create the PILFlavor class to generate the different license terms and export them.
- Mark the following methods deprecated, and they will be removed in the `v1.4.0`
   1. `LicenseClient.registerNonComSocialRemixingPIL`
   2. `LicenseClient.registerCommercialUsePIL`
   3.  `LicenseClient.registerCommercialRemixPIL`
   4. `LicenseClient.registerCreativeCommonsAttributionPIL`

## Related Issue
#576 
#579


gregsantos:
should we maintain both parameters with deprecation warnings for now?

// This is a breaking change that should be in a major version
royaltyPolicyAddress?: Address; // OLD
royaltyPolicy?: RoyaltyPolicyInput; // NEW

PR #629, opened 2025-09-15
feat: Re-design the registerDerivative method

bonnie57:
## Description
1. Add re-design the registerDerivative method 
2. Add unit and integration tests

### DIFF
--- a/packages/core-sdk/src/resources/ipAsset.ts
+++ b/packages/core-sdk/src/resources/ipAsset.ts
@@ -1715,9 +1715,10 @@ export class IPAssetClient {
   /**
    * Register a derivative IP asset, supporting both minted and mint-on-demand NFTs, with optional `derivData`, `royaltyShares` and `licenseTokenIds`, `maxRts`.
    *
-   * This method provides a unified interface for derivative IP registration with three main usage patterns:
+   * This method automatically selects and calls the appropriate workflow from 6 available methods based on your input parameters.
+   * Here are three common usage patterns:
    *
-   * **1. Derivative with License Terms and Royalty Distribution:**
+   * **1. Minted NFT with License Terms and Royalty Distribution:**
    * ```typescript
    * const result = await client.ipAsset.registerDerivativeIpAsset({
    *   nft: { type: "minted", nftContract: "0x...", tokenId: 1n },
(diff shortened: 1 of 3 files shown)


jacob-tucker (approved):

jacob-tucker:
Hi @bonnie57 !

I think it would be useful to provide default values for `maxMintingFee`, `maxRts`, and `maxRevenueShare`. This would improve dev ex and most of the time, user does not provide actual value for these. They just want defaults.

This is similar to how we provide default `licensingConfig` in `registerIpAsset` method if not provided.

jacob-tucker:
Another comment: Can you add `ipMetadata` when registering derivative? This should be supported, but I don't see it in the test

bonnie57:
> Hi @bonnie57 !
> 
> I think it would be useful to provide default values for `maxMintingFee`, `maxRts`, and `maxRevenueShare`. This would improve dev ex and most of the time, user does not provide actual value for these. They just want defaults.
> 
> This is similar to how we provide default `licensingConfig` in `registerIpAsset` method if not provided.

Hi @jacob-tucker, `maxMintingFee`, `maxRts`, and `maxRevenueShare` belong to `derivData`, which gives default values.

> Hi @bonnie57 !
> 
> I think it would be useful to provide default values for `maxMintingFee`, `maxRts`, and `maxRevenueShare`. This would improve dev ex and most of the time, user does not provide actual value for these. They just want defaults.
> 
> This is similar to how we provide default `licensingConfig` in `registerIpAsset` method if not provided.

Hi @jacob-tucker, I think your suggestion is a good idea. I have checked if  `maxMintingFee`, `maxRts`, and `maxRevenueShare` belong to `derivData`; [there is logic to give the default values](https://github.com/storyprotocol/sdk/blob/main/packages/core-sdk/src/utils/registrationUtils/registerValidation.ts#L137). But there are methods such as `registerIpAndMakeDerivativeWithLicenseTokens`, `mintAndRegisterIpAndMakeDerivativeWithLicenseTokens` accept `maxRts` args which don't support default values.
For the pr, it will finally call these methods; the giving default logic should put the responsive method instead of the `registerDerivativeIpAsset` method. So I will create a ticket to track two methods giving a default value about `maxRts`.

PR #633, opened 2025-09-17
feat: re-design the link Derivative method

bonnie57:
## Description
- Add the linkDerivative method
- Add unit and integration tests

## Breaking Change
Since `registerDerivative` and `registerDerivativeWithLicenseTokens` keep the same structure as `linkDerivative`,
to stay consistent, keep the same type of `LinkDerivativeResponse`
- Change the `registerDerivative`'s response from `RegisterDerivativeResponse` to `LinkDerivativeResponse`
- Change the `registerDerivativeWithLicenseTokens` response from `RegisterDerivativeWithLicenseTokensResponse` to `LinkDerivativeResponse`


bonnie57:
Hi @jacob-tucker, I agree with you. You are totally right.
Let me change it.

PR #663, opened 2025-11-18
feat: Auto approve license tokens 

bonnie57:
## Description
1. Add an `autoApproveLicenseTokens` to the following methods, and the default value is true:
- `registerIpAndMakeDerivativeWithLicenseTokens`
- `mintAndRegisterIpAndMakeDerivativeWithLicenseTokens`
2. Since these two methods are called by registerDerivativeIpAsset, also add the autoApproveLicenseTokens option to registerDerivativeIpAsset.
3. Add the integration and unit tests
4. Remove manually approved from the integration tests

### DIFF
--- a/packages/core-sdk/src/resources/ipAsset.ts
+++ b/packages/core-sdk/src/resources/ipAsset.ts
@@ -105,8 +105,8 @@ import {
 import { IpCreator, IpMetadata } from "../types/resources/ipMetadata";
 import { LicenseTerms } from "../types/resources/license";
 import { AggregateRegistrationRequest, SignatureMethodType } from "../types/utils/registerHelper";
-import { Erc20Spender } from "../types/utils/wip";
-import { calculateDerivativeMintingFee, calculateSPGWipMintFee } from "../utils/calculateMintFee";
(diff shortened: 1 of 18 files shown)


packages/core-sdk/src/resources/ipAsset.ts
```diff
@@ -2450,4 +2465,19 @@ export class IPAssetClient {
...
+    licenseTokenIds: LicenseTermsIdInput[],
+    autoApproveLicenseTokens?: boolean,
+  ): Promise<void> {
+    if (autoApproveLicenseTokens !== false) {
+      for (const licenseTokenId of licenseTokenIds) {
+        const txHash = await this.licenseTokenClient.approve({
+          to: this.derivativeWorkflowsClient.address,
+          tokenId: BigInt(licenseTokenId),
+        });
+        await waitTx(this.rpcClient, txHash);
```
jdubpark: Also, instead of waiting on one transaction at a time, can we send the transactions and wait all at once (using promises)?

lucas2brh (approved):

PR #704, opened 2026-03-06
fix: throw if address for chainId is missing

Ocheretovich:
Prevent getAddress from returning undefined when a mapping for the given chainId is missing. Adds a small fail-fast guard to throw an explicit error instead of returning an undefined address.

### DIFF
--- a/packages/wagmi-generator/resolveProxyContracts.ts
+++ b/packages/wagmi-generator/resolveProxyContracts.ts
@@ -23,7 +23,9 @@ function getAddress(
   if (typeof address == "string") {
     return address as Address;
   } else {
-    return address[chainId];
+    const resolved = address[chainId];
+    if (!resolved) throw new Error(`Address not found for chainId ${chainId}`);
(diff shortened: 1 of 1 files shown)


yingyangxu2026:
Thank you for your interest in our project. I'm curious why you're interested in this part of the code? This section of code is not exposed to users, and internally we will not map chainId to invalid addresses.

Ocheretovich:
While reviewing the code I noticed that address[chainId] could return undefined. Throwing an explicit error felt safer than silently returning an undefined value.

yingyangxu2026:
get it, I will communicate with other to decide whether merge the PR.

lucas2brh (approved):

yingyangxu2026 (approved):

yingyangxu2026:
@Ocheretovich  Commits must have verified signatures.

Ocheretovich:
@yingyangxu2026 signed and force-pushed the commit

